PR #253, opened 2025-11-04
yarn -> pnpm

sdqede:
### Key Changes
- Replace Lerna scripts with pnpm native commands
- Add custom workspace binary symlinking script for pnpm
- Fix Jest module resolution with `moduleNameMapper` for workspace dependencies
- Add missing devDependencies for TypeScript compilation
- Update build scripts and Jest snapshots

### Testing
✅ All local tests passing

### CI/CD
workflows already use pnpm


packages/cli/__tests__/__snapshots__/init.install.test.ts.snap
```diff
@@ -41,31 +41,31 @@ exports[`cmds:install - with initialized workspace and module installs a module
...
+  "../../extensions/@webql/base32/test/__snapshots__/base32.decode.test.ts.snap",
+  "../../extensions/@webql/base32/src/index.ts",
+  "../../extensions/@webql/base32/src/base32.ts",
+  "../../extensions/@webql/base32/sql/webql-base32--1.2.1.sql",
+  "../../extensions/@webql/base32/revert/schemas/base32/schema.sql",
+  "../../extensions/@webql/base32/revert/schemas/base32/procedures/encode.sql",
+  "../../extensions/@webql/base32/revert/schemas/base32/procedures/decode.sql",
+  "../../extensions/@webql/base32/deploy/schemas/base32/schema.sql",
+  "../../extensions/@webql/base32/deploy/schemas/base32/procedures/encode.sql",
+  "../../extensions/@webql/base32/deploy/schemas/base32/procedures/decode.sql",
```
pyramation, line 46: do you know why these got out of order?

packages/cli/jest.config.js
```diff
@@ -14,5 +14,11 @@ module.exports = {
   transformIgnorePatterns: [`/node_modules/*`],
   testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  modulePathIgnorePatterns: ['dist/*']
+  modulePathIgnorePatterns: ['dist/*'],
+  moduleNameMapper: {
+    '^@launchql/logger$': '<rootDir>/../../packages/logger/dist',
+    '^@launchql/(.*)$': '<rootDir>/../../packages/$1/dist',
+    '^pg-env$': '<rootDir>/../../packages/pg-env/dist',
+    '^pg-cache$': '<rootDir>/../../packages/pg-cache/dist',
```
pyramation: are these absolutely necessary?
Shouldn't this be picked up by pnpm workspace? 

I never had to do this in other packages, it seems that it could be cruft.

packages/introspectron/jest.config.js
```diff
@@ -14,5 +14,13 @@ module.exports = {
...
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  modulePathIgnorePatterns: ['dist/*']
+  modulePathIgnorePatterns: ['dist/*'],
+  moduleNameMapper: {
+    '^@launchql/logger$': '<rootDir>/../../packages/logger/dist',
+    '^@launchql/(.*)$': '<rootDir>/../../packages/$1/dist',
+    '^pgsql-test$': '<rootDir>/../../packages/pgsql-test/dist',
+    '^graphile-test$': '<rootDir>/../../packages/graphile-test/dist',
+    '^pg-env$': '<rootDir>/../../packages/pg-env/dist',
+    '^pg-cache$': '<rootDir>/../../packages/pg-cache/dist',
```
pyramation: same comment as before, this seems very very unmaintainble.

packages/launchql-gen/jest.config.js
```diff
@@ -14,5 +14,13 @@ module.exports = {
...
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  modulePathIgnorePatterns: ['dist/*']
+  modulePathIgnorePatterns: ['dist/*'],
+  moduleNameMapper: {
+    '^@launchql/logger$': '<rootDir>/../../packages/logger/dist',
+    '^@launchql/(.*)$': '<rootDir>/../../packages/$1/dist',
+    '^pgsql-test$': '<rootDir>/../../packages/pgsql-test/dist',
+    '^graphile-test$': '<rootDir>/../../packages/graphile-test/dist',
+    '^pg-env$': '<rootDir>/../../packages/pg-env/dist',
+    '^pg-cache$': '<rootDir>/../../packages/pg-cache/dist',
```
pyramation: this one also

maybe we use the `workspace:*` syntax and `pnpm` resolves this

packages/pg-codegen/jest.config.js
```diff
@@ -14,5 +14,13 @@ module.exports = {
...
-  modulePathIgnorePatterns: ['dist/*']
+  modulePathIgnorePatterns: ['dist/*'],
+  moduleNameMapper: {
+    '^@launchql/logger$': '<rootDir>/../../packages/logger/dist',
+    '^@launchql/(.*)$': '<rootDir>/../../packages/$1/dist',
+    '^pgsql-test$': '<rootDir>/../../packages/pgsql-test/dist',
+    '^graphile-test$': '<rootDir>/../../packages/graphile-test/dist',
+    '^pg-env$': '<rootDir>/../../packages/pg-env/dist',
+    '^pg-cache$': '<rootDir>/../../packages/pg-cache/dist',
+  },
```
pyramation: same here, we should NOT have `moduleNameMapper`

packages/pgsql-test/jest.config.js
```diff
@@ -14,5 +14,9 @@ module.exports = {
   transformIgnorePatterns: [`/node_modules/*`],
   testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  modulePathIgnorePatterns: ['dist/*']
+  modulePathIgnorePatterns: ['dist/*'],
+  moduleNameMapper: {
+    '^@launchql/logger$': '<rootDir>/../../packages/logger/dist',
+    '^@launchql/(.*)$': '<rootDir>/../../packages/$1/dist',
```
pyramation: this one too, can you explain what this was for?

perhaps, again, let's research if `workspace:*` syntax solves this instead of referencing the package versions directly.

packages/s3-streamer/__tests__/__snapshots__/uploads.test.ts.snap
```diff
@@ -1450,17 +1450,17 @@ exports[`uploads upload files via functions 1`] = `
   "md.md": {
     "contentType": "text/markdown",
     "contents": {
-      "etag": "2524b9c1caad97610a5149886a7e793f",
-      "sha": "62c8208c3f148ada7e1a1a3b8972ac5898dfbf89",
-      "uuid": "89250a3c-a794-5d26-8ee4-d692dff9fc35",
+      "etag": "9086b336429e90603656255396a5d462",
```
pyramation, line 1453: this snap getting updated is pretty serious. Do you know why these are changing? 

This needs to be documented, or likely reverted back.

packages/templatizer/src/generated/workspace.ts
```diff
@@ -44,10 +44,10 @@ export default [
   "scripts": {
     "clean": "lerna run clean",
     "build": "lerna run build --stream",
-    "build:dev": "lerna run build:dev --stream; yarn symlink",
+    "build:dev": "lerna run build:dev --stream; pnpm run symlink",
     "lint": "lerna run lint --parallel",
     "symlink": "symlink-workspace --logLevel error",
-    "postinstall": "yarn symlink"
+    "postinstall": "pnpm run symlink"
```
pyramation: we probably do not need to symlink anymore. Assuming we use `workspace:*` syntax, `pnpm` does the symlinking to simulate local modules

pyramation: `pnpm` workspaces should solve this

scripts/symlink-workspace.js
```diff
@@ -0,0 +1,41 @@
+const fs = require('fs');
+const path = require('path');
+
```
pyramation: this entire script should be deleted

packages/templatizer/src/generated/workspace.ts
yyyyaaa: yep, symlink is automatically run by pnpm when you do `pnpm install` anyways, this is unnecessary. It should be handled with `workspace:*` protocol

packages/cli/__tests__/__snapshots__/init.install.test.ts.snap
sdqede: I did some research, and found the order changed because glob.sync() in packages/cli/__tests__/init.install.test.ts doesn't guarantee file order—it depends on file system traversal, which differs between Yarn's flat node_modules and pnpm's symlinked structure.
I found out two options:
Option 1: Keep depth-first sort — matches yarn's apparent behavior but more complex
Option 2: Use alphabetical sort (.sort()) — deterministic but changes snapshot order
I prefer Option 2 for simplicity and reliability, but I'm happy to keep the original order if you prefer. Which do you prefer?

packages/cli/jest.config.js
sdqede: I tried workspace:^ and it worked, I will fix those moduleNameMapper issue

packages/introspectron/jest.config.js
sdqede: see reply above about workspace:^

packages/launchql-gen/jest.config.js
sdqede: see reply above about workspace:^

packages/pgsql-test/jest.config.js
sdqede: see reply above about workspace:^

packages/s3-streamer/__tests__/__snapshots__/uploads.test.ts.snap
sdqede: see reply above about glob.sync()

packages/templatizer/src/generated/workspace.ts
sdqede: "postinstall": "pnpm run symlink" deleted

scripts/symlink-workspace.js
sdqede: this file has been deleted

packages/pg-codegen/jest.config.js
sdqede: see reply above about workspace:^

packages/cli/__tests__/__snapshots__/init.install.test.ts.snap
pyramation: let's use option 2, I think you are right! :)

packages/s3-streamer/__tests__/__snapshots__/uploads.test.ts.snap
pyramation: so the hashes are just getting moved around, but the hashes themselves aren't changing?

PR #277, opened 2025-11-13
feature: fix server cors

Anmol1696:


packages/server/src/server.ts
```diff
@@ -14,7 +14,7 @@ import requestIp from 'request-ip';
 
 import { createApiMiddleware } from './middleware/api';
 import { createAuthenticateMiddleware } from './middleware/auth';
-import { cors } from './middleware/cors';
```
pyramation, line 17: removing this would break ability to use the sql version modules. Is there a way to do both, so we can temporarily get around the issue w/o breaking functionality?

packages/server/src/server.ts
```diff
@@ -14,7 +14,7 @@ import requestIp from 'request-ip';
...
+import { cors } from '@launchql/server-utils';
```
pyramation: if you read the `cors` from server-utils, you'll see that it doesn't have the cors modules (multi-tenant cors setup)

PR #309, opened 2025-11-16
[server] fix types for be build

pyramation:


packages/server/src/middleware/graphile.ts
```diff
@@ -1,19 +1,40 @@
...
 import { getPgPool } from 'pg-cache';
 import { postgraphile, PostGraphileOptions } from 'postgraphile';
 
-import PublicKeySignature, { PublicKeyChallengeConfig } from '../plugins/PublicKeySignature';
+import PublicKeySignature, {
+  PublicKeyChallengeConfig,
+} from '../plugins/PublicKeySignature';
+import { ApiStructure } from '../types';
+
+type GraphileRequest = Request & {
```
pyramation: let's do the same with this one

packages/server/src/middleware/auth.ts
```diff
@@ -1,20 +1,38 @@
...
+import { NextFunction, Request, RequestHandler, Response } from 'express';
 import { getPgPool } from 'pg-cache';
 import pgQueryContext from 'pg-query-context';
 
-export const createAuthenticateMiddleware = (opts: LaunchQLOptions): RequestHandler => {
+import { ApiStructure } from '../types';
 
-  return async (req: Request, res: Response, next: NextFunction): Promise<void> => {
-    const api = req.api;
+type AuthRequest = Request & {
```
pyramation: this too

PR #345, opened 2025-11-24
Added two options to generate schema from graphql

NorOldBurden:


packages/server/__tests__/schema.test.ts
```diff
@@ -0,0 +1,106 @@
+import { Pool } from 'pg';
+import express from 'express';
+import { postgraphile } from 'postgraphile';
+import { getGraphileSettings } from 'graphile-settings';
+import { buildSchemaSDL, fetchEndpointSchemaSDL } from '../src/schema';
+import { getPgPool } from 'pg-cache';
+
```
pyramation, line 9: can you meet with @theothersideofgod about graphile-test and launchql-test?

packages/server/src/schema.ts
```diff
@@ -0,0 +1,94 @@
...
+  schemas: string[];
+  graphile?: Partial<PostGraphileOptions>;
+};
+
+// Build GraphQL Schema SDL directly from Postgres using PostGraphile, without HTTP.
+export async function buildSchemaSDL(opts: BuildSchemaOptions): Promise<string> {
+  const database = opts.database ?? 'launchql'
+  const schemas = Array.isArray(opts.schemas) ? opts.schemas : []
+
+  const settings = getGraphileSettings({
```
pyramation, line 19: see if graphile-cache can help, maybe it will, maybe it will not.

pyramation: https://github.com/launchql/launchql/tree/main/packages/graphile-cache

packages/cli/__tests__/get-graphql-schema.e2e.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    expect(fs.existsSync(outDbFile)).toBe(true);
+    const sdl = fs.readFileSync(outDbFile, 'utf8');
+    expect(sdl).toContain('type Query');
+    expect(sdl).toContain('hello');
+  });
+
+  it('writes SDL to file when fetching from endpoint', async () => {
+    const argv: any = {
+      _: ['get-graphql-schema'],
+      endpoint: 'http://localhost:5555/graphql',
```
pyramation: is there a way to add `--headers`?

```
const token = process.env.MY_API_TOKEN; // or whatever your token source is

const response = await fetch("https://api.example.com/data", {
  method: "GET",
  headers: {
    "Authorization": `Bearer ${token}`,
    "Content-Type": "application/json",
  },
});
```

pyramation: maybe make an option for `--bearerToken`

or else we need something like this:

```
mycli fetch https://api.example.com \
  --header "X-Mode: fast" \
  --header "Authorization: Bearer 123" \
  --header "Content-Type: application/json"
```

pyramation:
```
mycli fetch https://1.2.3.4 --header "Host: example.com"
```

PR #371, opened 2025-12-04
Simplify lql init to rely on template questions, clean up version checks, and deduplicate CLI test helpers

sdqede:
# ✨ Summary

This PR:

- ✅ Makes `lql init` rely entirely on template `.questions.json` + **inquirerer** for identity questions (name, email, license, repo name, etc.).
- 🧹 Cleans up ALL‑CAPS keys and duplicated init logic in CLI tests and init flows.
- 🔀 Separates **pgpm** vs **@launchql/cli** update‑check state while sharing a common mechanism.
- 🧱 Improves template caching robustness and `create-gen-app` license template handling.

---

### `packages/pgpm/src/commands/init/workspace.ts`

Workspace init now:

- 📝 Prompts only for:
  - `name` (workspace name).
- 🚪 Closes the **Inquirerer** instance before `create-gen-app` runs to avoid double‑echo keystrokes.
- 📤 Delegates everything else (full name, email, license, `repoName`, `username`, etc.) to:
  - `pgpm-boilerplates/workspace/.questions.json`
  - inquirerer `defaultFrom` resolvers (e.g. `git.user.name`, `git.user.email`, `npm.whoami`).
- 🧹 Removes:
  - `getGitConfigInfo` usage.
  - `extraQuestions` and any special‑cased `repoName` / license questions.
  - Any **"LaunchQL User"** default.

### `packages/pgpm/src/commands/init/module.ts`

Module init now:

- 📝 Prompts only for:
  - `moduleName` (text).
  - `extensions` (checkbox, options from `getAvailableModules()`, with `allowCustomOptions: true`).
- 🚪 Closes **Inquirerer** before calling `initModule`.
- 📦 Passes a clean, camelCase answers object into `LaunchQLPackage.initModule`:
  - `moduleName`, `description`, `author`, `extensions`, `packageIdentifier`.
- 🧹 Removes all ALL‑CAPS answer keys and custom `username` / `email` logic; those now come solely from the module template’s `.questions.json`.

---

### `packages/core/src/core/template-scaffold.ts`

Template cache handling is now stricter and safer:

- ✅ Validates a cached template repo before use:
  - **Workspace templates** require:
    - `workspace/pgpm.json`
    - `workspace/package.json`
  - **Module templates** require:
    - `module/pgpm.plan`
    - `module/package.json`
- ♻️ If the cache is incomplete or invalid, we:
  - Clear that cache directory.
  - Reclones from `https://github.com/launchql/pgpm-boilerplates.git` using:

```bash
git clone --depth 1 --single-branch https://github.com/launchql/pgpm-boilerplates.git
```

- 🛡️ This prevents Jest (or a killed run) from leaving a broken cache that would previously cause:
  - `Not inside a workspace` errors.
  - Empty / inconsistent snapshots.

---

## 3️⃣ Update checks: shared logic, separate state

Shared logic for **pgpm** and **CLI** lives in:

- `packages/pgpm/src/utils/update-check.ts`
- `packages/pgpm/src/utils/update-config.ts`
- `packages/types/src/update.ts`

### 🔧 Config structure

```ts
interface UpdateCheckConfig {
  lastCheckedAt: number;
  latestKnownVersion: string;
  lastPromptedVersion?: string;
}
```

- ⏱ TTL: `UPDATE_CHECK_TTL_MS = 7 days`.

### 🔁 Behavior

- Read existing config and decide whether to hit npm (`shouldCheck`).
- On TTL expiry or missing config:
  - Fetch latest version from npm.
  - Write:

    ```json
    {
      "lastCheckedAt": <now>,
      "latestKnownVersion": "<npm-version>",
      "lastPromptedVersion": "<npm-version>" | undefined
    }
    ```

- When current version is outdated and `lastPromptedVersion !== latestKnownVersion`:
  - ⚠️ Log a one‑time upgrade warning.
  - Update config with `lastPromptedVersion = latestKnownVersion`.
- Otherwise, the config is left untouched.

### 📁 pgpm config path

From `packages/types/src/update.ts`:

```ts
export const UPDATE_CHECK_APPSTASH_KEY = 'pgpm-update-check';
export const UPDATE_PACKAGE_NAME = 'pgpm';
```

From `packages/pgpm/src/utils/update-config.ts`:

```ts
const toolName = options.toolName ?? 'pgpm';
const fileName = `${(options.key ?? UPDATE_CHECK_APPSTASH_KEY)
  .replace(/[^a-z0-9-_]/gi, '_')}.json`;
// → ~/.pgpm/config/pgpm-update-check.json
```

📌 Result:

- **pgpm** update config lives at:

```text
~/.pgpm/config/pgpm-update-check.json
```

### 📁 CLI (@launchql/cli) config path

From `packages/cli/src/commands.ts`:

```ts
const pkg = readAndParsePackageJson();
await checkForUpdates({
  command,
  pkgName: pkg.name,     // '@launchql/cli'
  pkgVersion: pkg.version,
  toolName: 'lql',       // directory namespace
  key: pkg.name          // '@launchql/cli'
});
```

`update-config.getConfigPath` normalizes the key:

```ts
const toolName = options.toolName ?? 'pgpm';
const fileName = `${(options.key ?? UPDATE_CHECK_APPSTASH_KEY)
  .replace(/[^a-z0-9-_]/gi, '_')}.json`;
```

- `toolName: 'lql'` → appstash dir → `~/.lql/config/`
- `key: '@launchql/cli'` → normalized to `_launchql_cli`

📌 Final path:

```text
~/.lql/config/_launchql_cli.json
```

### 🤔 Why the naming difference?

- **pgpm** is older and originally had a one‑off update check, so it used a fixed name: `'pgpm-update-check'`.
- We keep this to avoid breaking existing user configs in `~/.pgpm/config/pgpm-update-check.json`.
- For **@launchql/cli**, we intentionally key the config by npm package name:
  - Future‑proofs the helper so multiple CLIs can coexist without collisions.
  - Each package gets its own config file under its own `toolName` namespace (here, `lql`).

Under the hood, both use the same pattern:

```text
~/.<toolName>/config/<normalized key>.json
```

- **pgpm**: `toolName = 'pgpm'`, `key = 'pgpm-update-check'`
  - → `~/.pgpm/config/pgpm-update-check.json`
- **CLI**: `toolName = 'lql'`, `key = '@launchql/cli'`
  - → `~/.lql/config/_launchql_cli.json`

We can unify this later if we introduce a migration path, but for now this keeps existing pgpm configs working while giving the CLI a clean, package‑scoped cache.

---

### `packages/cli/test-utils/init-argv.ts`

New helpers:

```ts
export const addInitDefaults = (argv: ParsedArgs): ParsedArgs => {
  const baseName =
    (argv.moduleName as string) ||
    (argv.name as string) ||
    'module';

  const defaults = {
    fullName: 'Tester',
    email: 'tester@example.com',
    moduleName: argv.workspace ? 'starter-module' : baseName,
    username: 'tester',
    repoName: baseName,
    license: 'MIT',
    access: 'public',
    packageIdentifier: baseName,
    moduleDesc: baseName
  };

  return { ...defaults, ...argv };
};

export const withInitDefaults = (
  argv: ParsedArgs,
  defaultRepo = 'https://github.com/launchql/pgpm-boilerplates.git'
): ParsedArgs => {
  const args = addInitDefaults(argv);
  if (!Array.isArray(args._) || !args._.includes('init')) return args;

  return {
    ...args,
    repo: args.repo ?? defaultRepo,
    templatePath: args.templatePath ??
      (args.workspace ? 'workspace' : 'module')
  };
};
```

- Exported via `packages/cli/test-utils/index.ts`.

### `packages/cli/test-utils/fixtures.ts`

`TestFixture.runCmd` now:

```ts
if (Array.isArray(argv._) && argv._.includes('init')) {
  argv = withInitDefaults(argv, DEFAULT_REPO);
}
```

- Removes the inline defaults block and keeps all init defaults in one place.

### Tests using the helper

- `packages/cli/__tests__/init.test.ts`
  - Imports `withInitDefaults` from `../test-utils`.
  - Removes local `addDefaults` / `withInitDefaults` and the `DEFAULT_REPO` constant.
  - All init scenarios go through the shared helper.

- `packages/cli/__tests__/extensions.test.ts`
  - Uses `withInitDefaults(argv)` when running `init` as part of the extension flow.
  - No more ALL‑CAPS argv keys or repeated `||` chains in this test.

---

## 5️⃣ Boilerplates and license templates

- 🗂 The CLI no longer uses a local `./launchql/boilerplates` directory.
  - Both runtime and tests rely on the remote **pgpm-boilerplates** repo, aligning with `init-command.plan.md`.

### 📜 Workspace template license question

`pgpm-boilerplates/workspace/.questions.json` now includes the same license selection as the module template:

```json
{
  "name": "____license____",
  "message": "Choose a license",
  "type": "list",
  "options": ["MIT", "Apache-2.0", "ISC"],
  "required": true
}
```

### 📦 `dev-utils/packages/create-gen-app/package.json`

- Adjusted the copy script so license templates are copied into:

```text
dist/licenses-templates/
```

- This allows `create-gen-app`’s license resolver to find `APACHE-2.0.txt`, etc., and eliminates the:

> `License "Apache-2.0" is not supported`

warning when templates are present.

---

## ✅ Manual testing

All commands run locally from `./launchql` using `pnpm`.

### 🧪 CLI tests from a clean pgpm cache

```bash
rm -rf ~/.pgpm
pnpm --filter @launchql/cli test
```

Confirms:

- `pgpm` clones `https://github.com/launchql/pgpm-boilerplates.git` (with `--depth 1 --single-branch`) into the cache.
- All `@launchql/cli` Jest suites pass **without** requiring `--runInBand`.
- Cached repo is valid and reused on subsequent runs.

### 📖 CLI help

```bash
node packages/cli/dist/index.js init -h
```

Displays:

- `--workspace` flag for workspace init.
- `--repo`, `--template-path`, `--from-branch` options.

Matches README behavior:

- `lql init --workspace` → create workspace.
- `lql init` inside workspace → create module.

### 🔢 Version behavior and config

```bash
npm view @launchql/cli version
# confirm against packages/cli/dist/package.json

vim ~/.lql/config/_launchql_cli.json
cat ~/.lql/config/_launchql_cli.json
```

Example config:

```json
{
  "lastCheckedAt": 1764828956271,
  "latestKnownVersion": "4.17.17",
  "lastPromptedVersion": "4.17.17"
}
```

Verified:

- When `now - lastCheckedAt > 7 days`:
  - Next CLI invocation re‑checks npm, refreshes `latestKnownVersion`, and sets `lastCheckedAt = now`.
- When `latestKnownVersion > current CLI version` **and** `lastPromptedVersion` is different:
  - CLI prints a one‑time upgrade warning and updates `lastPromptedVersion` to the latest version (and `lastCheckedAt = now`).
- Otherwise, config stays untouched.

#### Workspace

```bash
rm -rf /tmp/lql-test
node packages/cli/dist/index.js init   --cwd /tmp/lql-test   --workspace
```

Creates `/tmp/lql-test/<workspace-name>/` with:

- `package.json`
- `pgpm.json`

All identity prompts (name, email, license, `repoName`, `username`, etc.) come from the workspace template’s `.questions.json` and inquirerer.

#### Module (inside workspace)

```bash
node packages/cli/dist/index.js init   --cwd /tmp/lql-test/<workspace-name>
```

Creates `/tmp/lql-test/<workspace-name>/packages/<module-name>/` with:

- `package.json`
- `pgpm.plan`

Extension choices and module metadata are entirely template‑driven; no extra prompts or git lookups in the CLI itself.


packages/cli/__tests__/extensions.test.ts
```diff
@@ -29,6 +33,24 @@ describe('cmds:extension', () => {
...
+      const defaults = {
+        fullName: 'Tester',
+        email: 'tester@example.com',
+        moduleName: argv.workspace ? 'starter-module' : argv.name || argv.MODULENAME || 'module',
+        username: 'tester',
+        repoName: (argv.name as string) || (argv.MODULENAME as string) || 'repo-name',
+        license: 'MIT',
+        access: 'public',
+        packageIdentifier: (argv.name as string) || (argv.MODULENAME as string) || 'module',
+        moduleDesc: (argv.name as string) || (argv.MODULENAME as string) || 'module'
```
pyramation: why ALL CAPS again? can't this just be simple? why so many `||`? please tame the AI ;)

pyramation: make it simple, just use very simple defaults and make code more readable

packages/cli/__tests__/init.test.ts
```diff
@@ -9,6 +12,29 @@ import { commands } from '../src/commands';
 import { setupTests, TestEnvironment, TestFixture } from '../test-utils';
 
 const beforeEachSetup = setupTests();
+const DEFAULT_REPO = 'https://github.com/launchql/pgpm-boilerplates.git';
+const addDefaults = (argv: ParsedArgs): ParsedArgs => {
+  const defaults = {
+    fullName: 'Tester',
+    email: 'tester@example.com',
+    moduleName: argv.workspace ? 'starter-module' : argv.name || argv.MODULENAME || 'module',
```
pyramation: again, same comment as before. THERE IS NO REASON FOR CAPS.

There is no reason to have all the `||`

packages/cli/__tests__/init.test.ts
```diff
@@ -100,13 +136,14 @@ describe('cmds:init', () => {
...
 
-      const argv: ParsedArgs = {
+      const argv: ParsedArgs = withInitDefaults({
         _: ['init'],
         cwd: fixture.tempDir,
         name: 'test-workspace-template',
         workspace: true,
-        templatePath: path.join(__dirname, '../../../boilerplates'),
-      };
+        templatePath: 'workspace',
```
pyramation: what is the difference between `workspace: true` and `templatePath: 'workspace'`?

should we just use `templatePath`?

packages/cli/__tests__/init.test.ts
```diff
@@ -147,14 +184,15 @@ describe('cmds:init', () => {
       });
 
       // Now initialize module with custom template
-      await commands({
+      await commands(withInitDefaults({
         _: ['init'],
         cwd: workspaceDir,
         name: 'test-module-template',
         MODULENAME: 'test-module-template',
```
pyramation: why `MODULENAME` that should not exist anymore.

packages/cli/test-utils/fixtures.ts
```diff
@@ -53,6 +54,25 @@ export class TestFixture {
...
 
+    // Default to local templates to avoid network and keep tests fast
+    const hasInitCommand = Array.isArray(argv._) && argv._.includes('init');
+    if (hasInitCommand) {
+      argv.repo = argv.repo ?? DEFAULT_REPO;
+      argv.templatePath = argv.templatePath ?? (argv.workspace ? 'workspace' : 'module');
+      const defaults = {
+        fullName: 'Tester',
+        email: 'tester@example.com',
+        moduleName: argv.workspace ? 'starter-module' : argv.name || argv.MODULENAME || 'module',
```
pyramation: this code looks copy and pasted.... let's deduplicate all 3 (or more) of the places this code is used, please.

packages/core/src/core/class/launchql.ts
```diff
@@ -420,18 +423,32 @@ export class LaunchQLPackage {
...
-    if (options.templateSource) {
-      const compiledTemplates = loadTemplates(options.templateSource, 'module');
-      templates = compiledTemplates.map((t: any) => t.render);
-    }
-    
-    writeRenderedTemplates(templates, targetPath, options);
+    const answers = {
+      ...options.answers,
+      name: options.name,
+      MODULEDESC: options.description,
```
pyramation: there is no reason at all for `MODULEDESC` to be all caps....

packages/pgpm/src/commands/init/module.ts
```diff
@@ -48,40 +46,74 @@ export default async function runModuleSetup(
   const answers = await prompter.prompt(argv, moduleQuestions);
   const modName = sluggify(answers.MODULENAME);
```
pyramation: why `MODULENAME` still

packages/pgpm/src/commands/init/module.ts
```diff
@@ -48,40 +46,74 @@ export default async function runModuleSetup(
...
 
+  if (!username || !email) {
```
pyramation: do we need this if statement? 

aren't these all just handled by inquirerer and the template?

packages/pgpm/src/commands/init/module.ts
```diff
@@ -48,40 +46,74 @@ export default async function runModuleSetup(
...
     ...argv,
     ...answers,
-    name: modName,
-    // @ts-ignore
     USERFULLNAME: username,
     USEREMAIL: email,
+    username,
+    email,
+    moduleName: modName,
+    MODULEDESC: answers.description || modName,
```
pyramation: remove this `ALL CAPS` stuff

packages/pgpm/src/commands/init/module.ts
```diff
@@ -48,40 +46,74 @@ export default async function runModuleSetup(
...
+  const templatePath = argv.templatePath as string | undefined;
 
-  project.initModule({
+  const templateAnswers = {
     ...argv,
     ...answers,
-    name: modName,
-    // @ts-ignore
     USERFULLNAME: username,
     USEREMAIL: email,
```
pyramation: more `CAPS`

USERFULLNAME
USEREMAIL

packages/pgpm/src/commands/init/workspace.ts
```diff
@@ -17,42 +15,104 @@ export default async function runWorkspaceSetup(
...
-    };
-    log.info(`Loading templates from local path: ${argv.templatePath}`);
-    const compiledTemplates = loadTemplates(source, 'workspace');
-    templates = compiledTemplates.map((t: any) => t.render);
+  const { cwd = process.cwd() } = argv;
+  const targetPath = path.join(cwd, sluggify(answers.name));
+
+  let { email, username } = getGitConfigInfo();
+
+  if (!username || !email) {
```
pyramation: why are we doing this? we should not need to ask the user for this?

pyramation: can't we just leave it to the inqurierer?

also, wasn't this a plugin that I wrote in inquirerer? why are we rebuilding all of this?

packages/pgpm/src/commands/init/workspace.ts
```diff
@@ -17,42 +15,104 @@ export default async function runWorkspaceSetup(
...
+      const identityAnswers = await prompter.prompt(argv, identityQuestions);
+      username = username || (identityAnswers as any).fullName;
+      email = email || (identityAnswers as any).userEmail;
+    }
+  }
+
+  // Collect any remaining identity values the templates need
+  const extraQuestions: Question[] = [];
+
+  // Repo name is used in README/badges; default to workspace folder name
```
pyramation: there should be ZERO knowledge of the variables in the templates.....

we should not have `extraQuestions`, only questions

packages/pgpm/src/commands/init/workspace.ts
```diff
@@ -17,42 +15,104 @@ export default async function runWorkspaceSetup(
...
+  if (!(argv as any).repoName) {
+    extraQuestions.push({
+      name: 'repoName',
```
pyramation: remove this

packages/pgpm/src/commands/init/workspace.ts
```diff
@@ -17,42 +15,104 @@ export default async function runWorkspaceSetup(
...
+      type: 'text',
+      default: sluggify(answers.name)
+    });
   }
 
-  writeRenderedTemplates(templates, targetPath, { ...argv, ...answers });
+  // License placeholder is required by templates
+  if (!(argv as any).license) {
+    extraQuestions.push({
+      name: 'license',
```
pyramation: remove this

packages/pgpm/src/commands/init/workspace.ts
```diff
@@ -17,42 +15,104 @@ export default async function runWorkspaceSetup(
...
+    outputDir: targetPath,
+    templateRepo,
+    branch: argv.fromBranch as string | undefined,
+    templatePath,
+    answers: {
+      ...argv,
+      ...answers,
+      ...extra,
+      workspaceName: answers.name,
+      fullName: username || email || 'LaunchQL User',
```
pyramation: replace any and all "LaunchQL User" — just put "user" if that. DO this everywhere you have this.

pyramation:
@sdqede we're getting close, much better. Still some work to do:

- remember, do NOT call getGitConfigInfo or try to get git info, let inqurierer handle that now with it's resolvers. https://github.com/launchql/pgpm-boilerplates/blob/968f5d6c7be0880252a6dce6d97837a486d683f9/workspace/.questions.json#L12

- there is way too much ALL CAPS stuff, remove it all. all of it.

- way to many extra questions or the creation of questions... I thought the entire point was to just parse the templates, and get the questions from .questions and the templates, and then ask user via inqurierer?  In other words, DO NOT create questions or try to guess any propery names like license, email, user, username, or ask for those. JUST USE THE TOOLS WE WROTE from create-gen-app

thank you @sdqede :)

packages/pgpm/src/commands/init/workspace.ts
sdqede: That logic is gone now. runWorkspaceSetup no longer calls getGitConfigInfo or asks for name/email itself – it only prompts for name, then hands control to create-gen-app/inquirerer + the template’s .questions.json, which resolve git defaults automatically.

packages/cli/__tests__/init.test.ts
sdqede: workspace: true selects the workspace init flow vs the default module init flow.
templatePath: 'workspace' only selects the subdirectory inside the template repo to render.
The user‑facing contract in the README already reflects this:
https://github.com/launchql/launchql/blob/main/packages/cli/README.md?plain=1#L27
Create a new workspace
lql init --workspace
Create a new module (run inside workspace)
lql init
So workspace: true is how we express “run the workspace init command” (matching lql init --workspace), and templatePath is just telling create-gen-app which folder (workspace/ vs module/) to use. That’s why the test keeps workspace: true and uses templatePath: 'workspace' only for template selection.

packages/cli/__tests__/extensions.test.ts
sdqede: In packages/cli/__tests__/extensions.test.ts I replaced the repeated name || moduleName || 'module' chains with a single baseName helper:
```
const baseName = (argv.moduleName as string) || (argv.name as string) || 'module';
const defaults = {
  fullName: 'Tester',
  email: 'tester@example.com',
  moduleName: argv.workspace ? 'starter-module' : baseName,
  username: 'tester',
  repoName: baseName,
  license: 'MIT',
  access: 'public',
  packageIdentifier: baseName,
  moduleDesc: baseName,
};
```
So we’re down to one || chain, all camelCase keys, and the defaults are much easier to read while keeping the same behavior.

packages/cli/__tests__/init.test.ts
sdqede: I’ve removed the ALL‑CAPS argv usage and the long || chains from both init.test.ts and extensions.test.ts.
For init.test.ts specifically, the defaults now look like:
```
const addDefaults = (argv: ParsedArgs): ParsedArgs => {
  const baseName = (argv.moduleName as string) || (argv.name as string) || 'module';
  const defaults = {
    fullName: 'Tester',
    email: 'tester@example.com',
    moduleName: argv.workspace ? 'starter-module' : baseName,
    username: 'tester',
    repoName: baseName,
    license: 'MIT',
    access: 'public',
    packageIdentifier: baseName,
    moduleDesc: baseName,
  };
  return { ...defaults, ...argv };
};

```
So:
all argv properties are camelCase (name, moduleName, etc.),
there is a single baseName fallback (moduleName → name → 'module'),
and the defaults are just simple assignments.

packages/cli/__tests__/init.test.ts
sdqede: that was from an earlier iteration. I’ve removed MODULENAME entirely; the tests now only use the camelCase moduleName (via a shared baseName helper), so this key no longer exists in the current code.

packages/cli/test-utils/fixtures.ts
sdqede: This logic was previously duplicated in several places.
I’ve consolidated the shared init defaults into:

packages/cli/test-utils/init-argv.ts
	•	addInitDefaults(argv)
Generates the common identity/module defaults—fullName, email, moduleName, username, repoName, license, etc.—derived from name / moduleName.
	•	withInitDefaults(argv, defaultRepo?)
Wraps addInitDefaults and additionally applies the repo and templatePath defaults used by init commands.

Updated Call Sites

All three locations now use the new helper:
	•	packages/cli/test-utils/fixtures.ts — TestFixture.runCmd
	•	packages/cli/__tests__/init.test.ts
	•	packages/cli/__tests__/extensions.test.ts

Result

This removes the previously duplicated setup blocks and centralizes the behavior in a single, reusable module.

packages/core/src/core/class/launchql.ts
sdqede: I’ve updated that code so it no longer uses MODULEDESC. The initModule answers now use camelCase keys:
```
const answers = {
  ...options.answers,
  name: options.name,
  moduleDesc: options.description,
  description: options.description,
  author: options.author,
  extensions: options.extensions
};
```
So the ALL‑CAPS field is gone and everything is consistent with the rest of the code

packages/pgpm/src/commands/init/module.ts
sdqede: All the ALL‑CAPS usages have been cleaned up

packages/pgpm/src/commands/init/module.ts
sdqede: I’ve removed that if (!username || !email) block entirely. Username and email are now fully handled by the template’s .questions.json plus inquirerer’s defaultFrom/validation, so the module init command no longer does any manual checks.

packages/pgpm/src/commands/init/module.ts
sdqede: I’ve removed the ALL‑CAPS fields from the module init flow. The answers we pass into initModule now use camelCase only (moduleName, description, etc.), and MODULEDESC is no longer referenced here

packages/pgpm/src/commands/init/module.ts
sdqede: I’ve removed those as well—USERFULLNAME and USEREMAIL no longer exist in this file (or anywhere in launchql). The module init flow now only uses camelCase fields and relies on the template + inquirerer for any user name/email handling

packages/pgpm/src/commands/init/workspace.ts
sdqede: I’ve removed the extraQuestions logic and any hard‑coded knowledge of template variables here.
The workspace init flow now only asks for the workspace name and then hands off entirely to create-gen-app/inquirerer, which reads all remaining questions from the template’s .questions.json

packages/pgpm/src/commands/init/workspace.ts
sdqede: I’ve removed that repoName / extraQuestions block from workspace.ts. The workspace init flow no longer tries to invent or default repoName itself—after asking for the workspace name, it hands off entirely to create-gen-app + the template’s .questions.json, which defines and resolves repoName (and other fields) on its own

packages/pgpm/src/commands/init/workspace.ts
sdqede: I’ve removed the custom license handling from workspace.ts as well. The workspace init command no longer pushes any extraQuestions; license selection now comes entirely from the template’s .questions.json (via create-gen-app/inquirerer), so the CLI doesn’t special‑case it anymore.

packages/pgpm/src/commands/init/workspace.ts
sdqede: That code is gone now too—there’s no LaunchQL User string anywhere in launchql (or in workspace.ts). The workspace init flow no longer sets a manual fullName default; name/email are handled entirely by the template questions and inquirerer

PR #392, opened 2025-12-12
Fixed minor issue for graphile import directory change after build. U…

NorOldBurden:
…pdated launchql-gen operation generation logics

### DIFF
--- a/graphile/graphile-plugin-connection-filter-postgis/src/index.ts
+++ b/graphile/graphile-plugin-connection-filter-postgis/src/index.ts
@@ -1,7 +1,14 @@
 import type { Plugin } from "graphile-build";
 
 import PostgisOperatorsPlugin from "./PgConnectionArgFilterPostgisOperatorsPlugin";
-import pkg from "../package.json";
+let pkg: any;
+try {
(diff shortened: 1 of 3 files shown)


graphile/graphile-plugin-connection-filter-postgis/src/index.ts
```diff
@@ -1,7 +1,14 @@
...
 import PostgisOperatorsPlugin from "./PgConnectionArgFilterPostgisOperatorsPlugin";
-import pkg from "../package.json";
+let pkg: any;
+try {
+  // In the built version (dist/), package.json is in the same directory
+  pkg = require("./package.json");
+} catch {
+  // In development (src/), package.json is one level up
+  pkg = require("../package.json");
+}
```
pyramation, line 11: this doesn't look very durable

pyramation: we should see how we're doing it elsewhere, esp. in our CLI

pyramation (changes requested):
let's look at what we're doing elsewhere, this code looks less durable than most

PR #412, opened 2025-12-17
Refactor/knative job send email link fn

Zetazzz:


jobs/job-utils/src/runtime.ts
```diff
@@ -139,3 +139,13 @@ export const getCallbackBaseUrl = (): string => {
   const port = getJobsCallbackPort();
   return `http://${host}:${port}/callback`;
 };
+
+// ---- Generic env helpers ----
+export const parseEnvBoolean = (
```
Anmol1696: PR is good, but this function is already defined in: `packages/env/src/env.ts`, can we please use that? so we dont have the expose the same function in 2 places.

PR #415, opened 2025-12-17
renew snapshorts

sdqede:
The cache TTL is 1 week. Here's what's happening:

Root Cause: The external boilerplate repo (pgpm-boilerplates) was updated to use jest --watchAll instead of jest --watch.

CI always gets a fresh clone (no cache), so it gets the new template
Locally i have a cached version from within the last week that still has jest --watch
Solutions:

Clear your local cache and re-run the test with -u:
```
pgpm cache clear
pnpm --filter @launchql/cli test -- -u
```


pyramation:
maybe this was already done

PR #531, opened 2025-12-29
Graphql server basic tests

Zetazzz:
added a constructive/graphql/server/__tests__/ for tests for graphql server.


graphql/server/__tests__/api-modules.test.ts
```diff
@@ -0,0 +1,627 @@
...
+};
+
+const authHeaders = {
+  Authorization: 'Bearer valid-token-123',
+};
+
+const metaDbExtensions = ['citext', 'uuid-ossp', 'unaccent', 'pgcrypto', 'hstore'];
+
+const getPgpmModulePath = (pkgName: string): string =>
+  dirname(require.resolve(`${pkgName}/pgpm.plan`));
```
pyramation, line 34: let's use `join(workspaceDir, 'pgpm.json')` — see `PgpmPackage`

pyramation: there should be a method on the PgpmPackage class?

pyramation: search for `export class PgpmPackage {` it lives in class/pgpm.ts

graphql/server/__tests__/api-modules.test.ts
```diff
@@ -0,0 +1,627 @@
...
+      database: db.config.database,
+    },
+    server: {
+      host: '127.0.0.1',
+      port: 0,
+      ...(strictAuth ? { strictAuth: true } : {}),
+    },
+    api: {
+      enableServicesApi: true,
+      isPublic: true,
```
Anmol1696, line 200: We need to test with is public to set to false and enable services set to true.
Note api.ts is actually broken so it should be failing in the test as well.

PR #538, opened 2025-12-31
WIP: feature: stronger types for jobs and functions

Anmol1696:
This can wait. We need to figure out a more generic types system across the board here, before getting into this.


functions/send-email-link/src/index.ts
```diff
@@ -46,6 +47,60 @@ const GetDatabaseInfo = gql`
   }
 `;
 
+type GetUserResult = {
```
Anmol1696, line 50: pretty sure we dont need to define these again here. need to look for defined types. 
Maybe see how dashboard does it??

Anmol1696:
Closing the PR, but the issue is not solved for. We need a shared library for functions to make typed queries to the graphql server.

PR #553, opened 2026-01-05
fix: wrong schema for getting metadata

yyyyaaa:
this PR adds 2 things to fix the server package, which in turns, will fix the cnc CLI command as well
Also adds some logs in DEV only for more visibility of a request


yyyyaaa:
closing in favor of a simpler PR

PR #575, opened 2026-01-07
feat: codegen source modes

yyyyaaa:
2 big changes:

- We dont even need to rely on `_meta` query anymore, as long as it's a postgraphile-based graphql endpoint or a graphql schema SDL, both sources work
- Redid the source -> generate pipeline, simplify and handling some special/outlier cases for latin words/names. 
- Added new CLI flag mode to use with plain graphql schema SDL `--schema` and just specify the schema file without needing to have a running server with an endpoint.

Everything works as before


graphql/codegen/src/cli/introspect/pluralize.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Latin plural suffixes that inflection handles differently than PostGraphile.
+ *
+ * The inflection library correctly singularizes Latin words (schemata → schematum),
+ * but PostGraphile uses English-style naming (schemata → schema).
+ *
+ * Format: [pluralSuffix, singularSuffix]
+ */
+const LATIN_SUFFIX_OVERRIDES: Array<[string, string]> = [
```
pyramation, line 16: let's make our own inflection library

PR #613, opened 2026-01-14
(WIP) fix(codegen): missing generic types for react query mode hooks

yyyyaaa:
1. Missing generic type parameters on execute() calls - The execute<TData, TVariables>() function was called without type parameters, causing TypeScript to infer unknown for return types.
  2. Missing table type imports in mutations - When a table field references another table type (e.g., schema?: Schema), the referenced type wasn't imported, causing "Cannot find name" errors.

  Solution

  - Added createTypedCallExpression() helper to generate properly typed execute<Result, Variables>() calls
  - Updated mutation generators to track and import all table types used in scalar fields
  - Added list filter types (StringListFilter, IntListFilter, UUIDListFilter) to ORM generator

### DIFF
--- a/graphql/codegen/examples/orm-sdk.ts
+++ b/graphql/codegen/examples/orm-sdk.ts
@@ -5,7 +5,7 @@
 import {
   createClient,
   GraphQLRequestError,
-} from '../examples/output/generated-orm';
+} from './output/generated-orm';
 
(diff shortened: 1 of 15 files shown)


yyyyaaa:
I'm discovering some more issues, don't merge yet plz

PR #615, opened 2026-01-16
Update mode to support both react query and orm generation for codegen. Added header support in cli.

NorOldBurden:


graphql/server/src/middleware/gql.ts
```diff
@@ -19,7 +19,7 @@ const apiSelect = {
     select: { schemaName: true },
     connection: true,
   },
-  schemasByApiSchemaApiIdAndSchemaId: {
+  schemataByApiSchemaApiIdAndSchemaId: {
```
pyramation: this looks like the OLD naming!

yyyyaaa:
I have a few questions:
The PR could have simply been:
  1. Add headers to GenerateOptions interface
  2. Pass options.headers through to fetchSchema()
  3. Add CLI --header parsing
Why did we remove fetch here? `fetch` is supported anywhere and simply works.

Was there a specific bug with fetch (SSL issues, proxy problems, Node version compatibility)?

yyyyaaa:
Was Host header override the specific issue you hit? Can you share the error or scenario? I need to know the exact scenario that's not working that you run into here. Plz provide some steps to reproduce

NorOldBurden:
> Was Host header override the specific issue you hit? Can you share the error or scenario? I need to know the exact scenario that's not working that you run into here. Plz provide some steps to reproduce

As I showed back in the meeting
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header
The fetch doesn't support for the list forbidden list override where host is one of them.

packages/cli/src/commands/codegen.ts
```diff
@@ -60,24 +77,50 @@ export default async (
     : getEnvOptions();
   const schemasArg = (argv.schemas as string) || '';
 
+  // Parse repeatable --header args into a headers object
```
yyyyaaa: this whole thing could be a separate function, hard to read like this imo

yyyyaaa (changes requested):
and also what's with the code changes for switching from `.forEach` into `for()` loops?
Is that intentional or AI hallucination?

packages/cli/src/commands/codegen.ts
NorOldBurden: It was just a lint issue, it keeps showing red line. So you prefer the old style?

NorOldBurden: I moved headers logics into a function, also reformat the foreach to keep the original syntax while cast variable into any to avoid lint issue.

PR #621, opened 2026-01-19
applied makage and standardize the pkg in jobs and fns

Zetazzz:


jobs/knative-job-worker/__tests__/req.test.ts
```diff
@@ -17,16 +17,16 @@ describe('knative request wrapper', () => {
     process.env.PGPORT = '5432';
     process.env.PGDATABASE = 'jobs';
     process.env.JOBS_SCHEMA = 'app_jobs';
-    process.env.INTERNAL_JOBS_CALLBACK_URL =
+    process.env.JOBS_CALLBACK_BASE_URL =
```
Anmol1696: are we changing the env variables?? this will start breaking what we have in the cloud. 
If we change anything here, can we outline it properly in the PR so we can make related changes in the cloud PR

Zetazzz: Ah, sorry, I think this's confused by the JOBS_CALLBACK_BASE_URL usage in getCallbackBaseUrl like:
export const getCallbackBaseUrl = (): string => {
  if (process.env.JOBS_CALLBACK_BASE_URL) {
    return process.env.JOBS_CALLBACK_BASE_URL;
  }
  const host = process.env.JOBS_CALLBACK_HOST || 'jobs-callback';
  const port = getJobsCallbackPort();
  return `http://${host}:${port}/callback`;
};

I think I should change JOBS_CALLBACK_BASE_URL here to INTERNAL_JOBS_CALLBACK_URL

PR #622, opened 2026-01-19
fix(codegen): use correct deleted{Entity}NodeId field in delete mutation

yyyyaaa:
**Problem**

  The React Query hooks generator was producing delete mutations with a hardcoded deletedId field, but our backend convention switched to use type-specific field names. I think this is very recent?
  
  `deleted{EntityName}NodeId`

  # Before (broken)
  `deleteUser(input: $input) { deletedId }`

  # After (correct)
  `deleteUser(input: $input) { deletedUserNodeId }`

  **Solution**

  - Added getDeletedNodeIdFieldName() helper that infers the correct field name from the TypeRegistry (schema introspection), with fallback to PostGraphile naming convention
  - Updated buildDeleteMutationAST() and generateDeleteMutationHook() to use the inferred field name
  - Passed TypeRegistry from the codegen orchestrator to mutation generators


pyramation:
wait, what is `deletedUserNodeId` for? did the `Node` plugin get re-enabled?

pyramation:
I think this is an incorrect fix, I'm not sure we should be referencing anything `NodeId` or close to it

https://github.com/constructive-io/constructive/blob/a35d4d29df0e279cd559454f100e723b0fc31ebf/graphile/graphile-settings/src/index.ts#L90

we purposefully skip the `NodePlugin` 

Node ids are a graphql thing we don't use and they're literally useless. Maybe something broke, or we're not using graphile-settings somewhere in codegen?

https://www.graphile.org/postgraphile/node-id/

yyyyaaa:
I don't know about enabling it, but this is what I see when I run latest constructive db

Screenshot 2026-01-20 at 08 05 04

yyyyaaa:
literally everything that `delete` something mutation has it

pyramation:
From research:

The deletedDatabaseNodeId field is *not a bug* - it's expected behavior even with NodePlugin disabled.

*What's happening:*

In <ref_snippet file="/home/ubuntu/repos/constructive/graphile/graphile-settings/src/index.ts" lines="90-90" />, the NodePlugin is correctly being skipped:
typescript
skipPlugins: [NodePlugin],

However, the deletedDatabaseNodeId field comes from a *different mechanism*. PostGraphile's mutation plugins (specifically PgMutationUpdateDeletePlugin) add deleted<TypeName>NodeId fields to delete mutation payloads for Relay-style cache invalidation - this is separate from the NodePlugin.

The field name itself is generated by the deletedNodeId inflector in <ref_snippet file="/home/ubuntu/repos/constructive/graphile/graphile-simple-inflector/src/index.ts" lines="211-215" />:
typescript
deletedNodeId(this: PgInflection, table: PgClass) {
  return this.camelCase(
    `deleted-${this.singularize(table.name)}-${nodeIdFieldName}`
  );
}

*What is NodeId?*

NodeId is PostGraphile's implementation of Relay's Global Object Identification spec - a globally unique, opaque ID that encodes the table name and primary key. It's used for:
1. The node(nodeId: ID!) query to fetch any object by its global ID
2. Cache invalidation in Relay clients (the deleted<Type>NodeId field)

*Skipping NodePlugin removes:*
- The nodeId field on types
- The node query

*But it does NOT remove:*
- The deleted<Type>NodeId field on delete mutation payloads (added by mutation plugins)

If you want to remove deletedDatabaseNodeId, you'd need a custom plugin to omit that field from mutation payloads, or use smart tags to omit it.

yyyyaaa:
removed the `deleted<entity>Id` returning types altogether, kept `clientMutationId` by default, which everything has

yyyyaaa:
1 test failing, fixing

PR #616, opened 2026-01-16
feature: update function to use admin server with headers

Anmol1696:
## Summary
Enable `send-email-link` function to work with jobs infrastructure via header-based GraphQL routing.

## Changes
- Fix schema field name mismatch in API middleware (`schemaNamesFromExt` → `apiExtensions`)
- Fix GraphQL response structure for `apiByDatabaseIdAndName` query
- Add header-based routing support (`X-Database-Id`, `X-Api-Name`, etc.)
- Strip subdomain for localhost domains (generates `http://localhost:3000` not `http://app.localhost:3000`)
- Add SMTP email support option
- Update docker-compose with admin/public server split
- Replace console.log with structured logger

## Testing
- [x] Existing e2e tests pass
- [x] Added tests for `forgot_password` and `email_verification` email types
- [x] Manual docker-compose verification

## Breaking Changes
- `simple-email` function removed from docker-compose (use `send-email-link` instead)
- Network renamed to `constructive-tests-net`
- Admin server port changed to 3101

PR #630, opened 2026-01-21
[graphql/server] fix/api routing headers bugs

yyyyaaa:
- fix api routing with custom headers
- consolidate types
- use codegen ORM without the plain buildDocument api
- re-run codegen


yyyyaaa:
@pyramation FYI your fixes regarding moving to prettier made the codegen with --endpoint option not usable within any other directory, the `codegen-prettier.json` file is not bundled/not in the `dist` of the codegen dependency

yyyyaaa:
I'm fixing that too in this PR if you don't mind, in a separate commit, and all commits will be rebased without squashing

graphql/server/src/middleware/api.ts
```diff
@@ -11,61 +11,50 @@ import { getPgPool } from 'pg-cache';
...
-      api.apiModules?.nodes?.map((node) => ({
-        name: node.name,
-        data: node.data,
-      })) || [],
-    rlsModule: api.rlsModule,
-    domains,
-    databaseId: api.databaseId,
-    isPublic: api.isPublic,
+type ApiOptions = PgpmOptions & {
+  api?: {
```
Anmol1696: this seems be an inline type, maybe not correct to put here.

graphql/server/src/middleware/api.ts
```diff
@@ -11,61 +11,50 @@ import { getPgPool } from 'pg-cache';
...
+    exposedSchemas?: string[];
+    anonRole?: string;
+    roleName?: string;
+    defaultDatabaseId?: string;
+    metaSchemas?: string[];
+    isPublic?: boolean;
   };
 };
 
+type ApiError = { errorHtml: string };
```
Anmol1696: maybe better to move it into types instead of inline types.

yyyyaaa: yep, agree, let me add another commit to clean up

graphql/server/src/middleware/api.ts
```diff
@@ -156,155 +165,140 @@ const getHardCodedSchemata = ({
...
-        schemaNames: { nodes: [] as Array<{ schemaName: string }> },
-        apiModules: [] as Array<any>,
-      },
-    },
+}): ApiStructure => {
+  const api: ApiStructure = {
+    dbname: opts.pg?.database ?? '',
+    anonRole: 'administrator',
+    roleName: 'administrator',
+    schema: schemata,
```
Anmol1696, line 167: is this correct?

yyyyaaa: yes, it's correct

Anmol1696:
@pyramation could you have a look, this works locally, and for my branch as well: https://github.com/constructive-io/constructive/pull/616.

graphql/server/src/middleware/api.ts
yyyyaaa: `getHardCodedSchemata` and ` getMetaSchema` are actually pretty similar, only differs in 1 param, should we refactor them and combine? Not sure

graphql/server/src/middleware/api.ts
```diff
@@ -74,21 +53,41 @@ const getPortFromRequest = (req: Request): string | null => {
   return parts.length === 2 ? `:${parts[1]}` : null;
 };
 
+const parseSchemataHeader = (schemata: string): string[] =>
```
pyramation, line 56: schemata? are we still using schemata?

PR #676, opened 2026-01-29
feat(codegen): add getHeaders callback for dynamic request headers

yyyyaaa:


graphql/codegen/package.json
```diff
@@ -55,6 +55,7 @@
   "dependencies": {
     "@0no-co/graphql.web": "^1.1.2",
     "@babel/generator": "^7.28.6",
+    "@babel/parser": "^7.28.6",
```
pyramation: let's remove this?

PR #700, opened 2026-02-10
(WIP) Feat/codegen v5 on develop v5

yyyyaaa:


yyyyaaa:
there's a big TS issue with nested select clause that I'm solving but generated output looks good and compiling

PR #703, opened 2026-02-11
fix(graphile-settings): standalone UppercaseEnumsPlugin for v4 CONSTANT_CASE enum values

pyramation:
# fix(graphile-settings): add UppercaseEnumsPlugin for v4 CONSTANT_CASE enums

## Summary

Adds a standalone `UppercaseEnumsPlugin` that overrides PostGraphile v5's `enumValue` inflector to uppercase enum values, restoring the v4 behavior where PostgreSQL enum values like `app`, `core`, `module` become `APP`, `CORE`, `MODULE` in the GraphQL schema.

In v5, the default `enumValue` inflector in `PgCodecsPlugin` just calls `coerceToGraphQLName(value)`, which preserves the original Postgres casing. This plugin delegates to the previous inflector (retaining special character handling) then applies `.toUpperCase()`.

Three files changed:
- **New:** `plugins/uppercase-enums.ts` — standalone plugin + preset
- **Modified:** `plugins/index.ts` — re-exports the new plugin
- **Modified:** `presets/constructive-preset.ts` — adds `UppercaseEnumsPreset` to the main preset

This is an alternative to [PR #702](https://github.com/constructive-io/constructive/pull/702), which inlined the same logic into `InflektPlugin`. This approach keeps it as a separate, independently toggleable plugin.

## Review & Testing Checklist for Human

- [ ] **Verify behavior with a live schema**: No integration test exists — spin up PostGraphile with a custom enum type (e.g., `CREATE TYPE module_type AS ENUM ('app', 'core', 'module')`) and confirm the GraphQL schema shows `APP`, `CORE`, `MODULE`
- [ ] **Check impact on non-custom enums**: `.toUpperCase()` is applied to *all* enum values (OrderBy enums like `ID_ASC`, filter enums, etc.). These should already be uppercase so it should be a no-op, but worth spot-checking that nothing breaks
- [ ] **Confirm v4 parity for mixed-case enums**: If any Postgres enums use camelCase values like `myValue`, this produces `MYVALUE` (not `MY_VALUE`). Verify this matches v4's behavior — v4 also did a simple uppercase, not a CONSTANT_CASE transform

### Notes
- The `previous!` non-null assertion is safe because `PgCodecsPlugin` always defines `enumValue`, but if the plugin ordering ever changes this could throw
- Link to Devin run: https://app.devin.ai/sessions/9842845c47f54bdeb5a1aed39e1f53e7
- Requested by: @pyramation

### DIFF
--- a/graphile/graphile-settings/src/plugins/index.ts
+++ b/graphile/graphile-settings/src/plugins/index.ts
@@ -63,6 +63,12 @@ export {
 } from './pg-type-mappings';
 export type { TypeMapping } from './pg-type-mappings';
 
+// Uppercase enum values to match v4 CONSTANT_CASE convention
+export {
+  UppercaseEnumsPlugin,
+  UppercaseEnumsPreset,
+} from './uppercase-enums';
+
 // Search plugin for tsvector full-text search conditions (includes TsvectorCodec)
 export {
   PgSearchPlugin,
--- a/graphile/graphile-settings/src/plugins/uppercase-enums.ts
+++ b/graphile/graphile-settings/src/plugins/uppercase-enums.ts
@@ -1,0 +1,33 @@
+import type { GraphileConfig } from 'graphile-config';
+
+/**
+ * Plugin that uppercases PostgreSQL enum values in the GraphQL schema.
+ *
+ * WHY THIS EXISTS:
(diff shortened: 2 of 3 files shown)


Zetazzz:
The 702's been merged, we close this for now

PR #717, opened 2026-02-13
refactor(graphql-codegen): handle pg resource and cleanup

yyyyaaa:


yyyyaaa:
Repurposed the PR for a light refactor change as requested

PR #826, opened 2026-03-16
Feat/img process

Zetazzz:


pyramation:
converting this to a draft for now until it uses our standard tooling

PR #849, opened 2026-03-19
fix(send-email-link): use X-Meta-Schema for platform data queries

theothersideofgod:
## Problem

send-email-link fails for newly provisioned databases (e.g., kanban, crm) with the error:

  Cannot query field "databases" on type "Query"

However, it works correctly for the `constructive` database.

## Root Cause

send-email-link queries platform-level data (databases, sites, domains, siteThemes, siteModules) using `X-Api-Name: private` header. This relies on the database's `private` API having access to meta schemas.

The issue is:
- `constructive` database: private API includes meta schemas (seed data)
- New databases (kanban, crm): private API does NOT include meta schemas

The `private` API's accessible schemas are determined by the `services_public.api_schemas` table. For `constructive`, this was manually configured to include `metaschema_public` and `services_public`. But the provision flow does not add these meta schemas for new databases.

## Analysis

GetDatabaseInfo query needs:
- `databases` table (metaschema_public)
- `sites`, `domains`, `siteThemes`, `siteModules` tables (services_public)

These are platform-level data, not per-database business data. They should be accessed via `X-Meta-Schema: true` header, which directly exposes the meta schemas regardless of which database is being queried.

## Solution

Change the `meta` GraphQL client to use `useMetaSchema: true` instead
of `X-Api-Name: private`. This ensures:

1. Platform data is always accessible via X-Meta-Schema header
2. No dependency on per-database API configuration
3. Works for all databases (existing and new)

The `client` (for GetUser query) still uses `X-Api-Name: private` since it needs per-database user data.

## Before

```typescript
const meta = createGraphQLClient(metaGraphqlUrl, {
  databaseId,
  ...(apiName && { apiName }),  // X-Api-Name: private
});
```

## After

```typescript
const meta = createGraphQLClient(metaGraphqlUrl, {
  useMetaSchema: true,  // X-Meta-Schema: true
});
```

## Testing

Verified with:
- constructive database: ✅ email sent successfully
- kanban database: ✅ email sent successfully (was failing before)


NorOldBurden (approved):

Zetazzz (commented):
no problem

Zetazzz (approved):

PR #937, opened 2026-03-30
fix: enable orderBy on all columns alongside filterBy in v5 plugin

yyyyaaa:
The problem: PostGraphile v5's PgIndexBehaviorsPlugin adds both -filterBy and -orderBy to non-indexed columns. The existing
  EnableAllFilterColumnsPlugin only restored filterBy — it missed re-enabling orderBy.

  The fix (this diff): Adds 'orderBy' alongside 'filterBy' in the behavior callback (line 77):
  // Before
  return [behavior, 'filterBy'];
  // After
  return [behavior, 'filterBy', 'orderBy'];

  Impact: Without this, any column that doesn't have a database index would be missing from the orderBy enum in GraphQL
  connections — meaning clients couldn't sort by those columns. This is a significant gap since the plugin's intent was always
  to remove v5's index-only restriction.


yyyyaaa:
this is needed for the sorting by column features in dashboard side:

https://github.com/user-attachments/assets/18d72c28-aba7-42ff-b2f5-328dd81a53ff

PR #1025, opened 2026-04-24
feat(codegen): bake isomorphic fetch into ORM client template (#754)

yyyyaaa:
Generated SDKs now handle Node's *.localhost DNS and Host-header quirks out of the box, so no separate @constructive-io/node shim is required. The template resolves a default fetch at runtime: browsers/Deno/Bun use globalThis.fetch; Node wraps it with a node:http/node:https implementation that rewrites *.localhost hostnames to plain localhost (RFC 6761 — resolves on both IPv4 and IPv6 loopback) while preserving the original Host header so server-side subdomain routing keeps working.

@constructive-io/node is now a thin deprecation shim over @constructive-io/sdk for backwards compatibility. Also ports the codegen enum-type collection fix so input types referenced only by ENUM arguments aren't dropped.

Validated end-to-end against a live PostGraphile dev server: sign-in at auth.localhost:3000 returns a Bearer token that unlocks authenticated queries against api.localhost:3000 — all handled by the generated SDK's baked-in FetchAdapter with zero caller-side configuration.


yyyyaaa:
Core fix — single file does all the work
  - graphql/codegen/src/core/codegen/templates/orm-client.ts (+208) — bakes isomorphic FetchAdapter into every generated ORM
  client: runtime detection (browser/Node), Node-only dynamic import of node:http/node:https, *.localhost → localhost rewrite,
  Host header preservation

  Codegen enum-type collection fix (ported from earlier work)
  - graphql/codegen/src/core/codegen/orm/input-types-generator.ts (+14)
  - graphql/codegen/src/core/codegen/orm/custom-ops-generator.ts (+17)

  Codegen regression tests
  - graphql/codegen/src/__tests__/codegen/client-generator.test.ts (+24) — two new cases pinning the fetch?: injection point
  and the Node quirks handling
  - graphql/codegen/src/__tests__/codegen/input-types-generator.test.ts (+33) — enum collection coverage

  @constructive-io/node deprecation
  - sdk/constructive-node/src/fetch.ts (new, +133) — standalone Node fetch helper for callers not on codegen 5.x yet
  - sdk/constructive-node/src/__tests__/fetch.test.ts (new, +89) — echo-server test for *.localhost rewrite + Host preservation
  - sdk/constructive-node/src/node-http-adapter.ts (-167 net) — slimmed to delegate to ./fetch
  - sdk/constructive-node/src/index.ts (~35) — re-export @constructive-io/sdk + deprecation JSDoc
  - sdk/constructive-node/README.md (~41) — deprecation warning + migration diff
  - sdk/constructive-node/package.json (~2) — description flipped to "DEPRECATED"
  - sdk/constructive-node/jest.config.js (new, +18) — standard ts-jest setup

  Mechanically generated (~1688 lines, 9 files)

  - sdk/constructive-sdk/src/{admin,auth,objects,public}/orm/client.ts (4 × +185) — pnpm --filter @constructive-io/sdk generate
  - sdk/constructive-react/src/{admin,auth,objects,public}/orm/client.ts (4 × +185) — pnpm --filter @constructive-io/react
  generate
  - graphql/codegen/src/__tests__/codegen/__snapshots__/client-generator.test.ts.snap (+208) — jest -u after the template
  change

NorOldBurden (approved):

yyyyaaa:
@pyramation if you don't mind, could you use the generated sdk (not the `node` one, just `sdk`) for a final local test if you don't mind. Although Gefei tested locally and all works fine but just to be sure :D

pyramation:
thanks! closed since we don't actually need isomorphic fetch, and then bundled this into `@constructive-io/fetch`, thanks dude~

yyyyaaa:
oops! bummer, at least it gets rolled into a package. Maybe misunderstanding on my end but didn't we say that isomorphic fetch is needed and then get rid of the node sdk version?

PR #1026, opened 2026-04-25
fix: orm interfaces findX

yyyyaaa:
fixes https://github.com/constructive-io/constructive-planning/issues/752

ORM model interfaces (the core fix)

  findFirst — single object return ✓
  findFirst<S extends UserSelect>(
    args: FindFirstArgs<S, UserFilter, UserOrderBy> & { select: S } & StrictSelect<S, UserSelect>
  ): QueryBuilder<{
    user: InferSelectResult<UserWithRelations, S> | null;   // ← singular, was array
  }>
  With transform callback: (data) => ({ user: data.users?.nodes?.[0] ?? null })

  findOne — single object return ✓
  Same { user: ... | null } shape as findFirst, using buildFindManyDocument with where: { id: { equalTo: args.id } }, first: 1.

  Coverage across SDKs (constructive-sdk only — others mirror)

  - findFirst<...> declarations: 168 (auth: 9, objects: 5, admin: 34, public: 120)
  - Unique singular { xxx: InferSelectResult<XxxWithRelations, S> \| null } return shapes: 120 distinct entities
  - Regression check: 0 instances of findFirst returning an array

  FindFirstArgs interface (in generated select-types.ts)

  export interface FindFirstArgs<TSelect, TWhere, TOrderBy = never> {
    select?: TSelect;
    where?: TWhere;
    orderBy?: TOrderBy[];   // ← new orderBy support
  }

  CLI (sdk/constructive-cli)

  - Help text includes --orderBy under "Find-First Options" (verified user.ts and others)
  - handleFindFirst calls parseFindFirstArgs<FindFirstArgs<UserSelect, UserFilter, UserOrderBy>>(argv, defaultSelect)

  React hooks (sdk/constructive-react)

  - useUserQuery returns UseQueryResult<{ user: InferSelectResult<UserWithRelations, S> \| null }>
  - Delegates to getClient().user.findOne(...) — picks up the new singular shape transparently
  - fetchUserQuery and prefetchUserQuery mirror the same contract

  migrate-client SDK

  Same pattern verified on SqlAction.findFirst — returns { sqlAction: ... | null } with SqlActionOrderBy type param.

  Build + typecheck (already done)

  - All 4 SDKs build clean
  - tsc --noEmit passes with zero errors on all 4 SDKs
  - 306 codegen tests pass, 119 snapshots green


pyramation:
nice is this ready to review?

yyyyaaa:
yep, pr ready

pyramation:
out of curiosity, breaking changes? so we'll need to change FE before we do this one?

graphql/query/src/runtime/index.ts
```diff
@@ -24,5 +24,5 @@ export { parseType, print } from '@0no-co/graphql.web';
 export type { GraphQLAdapter, GraphQLError, QueryResult } from '@constructive-io/graphql-types';
 
 // Isomorphic fetch with *.localhost DNS + Host header fix for Node.js
-export { createFetch } from './localhost-fetch';
-export type { FetchFunction } from './localhost-fetch';
+export { createFetch } from './localhost-fetch.js';
+export type { FetchFunction } from './localhost-fetch.js';
```
pyramation: why `.js`? we are not using esm

graphql/query/src/runtime/localhost-fetch.ts
```diff
@@ -122,12 +134,13 @@ export function createFetch(): FetchFunction {
...
-      // eslint-disable-next-line @typescript-eslint/no-require-imports
-      const http = require('node:http');
-      // eslint-disable-next-line @typescript-eslint/no-require-imports
-      const https = require('node:https');
-      _fetch = buildNodeFetch(http, https);
-      return _fetch;
+      const nodeRequire = getNodeModuleLoader();
+      if (nodeRequire) {
+        const http = nodeRequire('node:' + 'http') as typeof import('node:http');
+        const https = nodeRequire('node:' + 'https') as typeof import('node:https');
```
pyramation: 👀

graphql/query/package.json
```diff
@@ -6,6 +6,18 @@
...
+      "types": "./index.d.ts",
+      "import": "./esm/index.js",
+      "require": "./index.js"
+    },
+    "./runtime": {
+      "types": "./runtime/index.d.ts",
+      "import": "./esm/runtime/index.js",
+      "require": "./runtime/index.js"
+    }
+  },
```
pyramation: yea we don't use exports — see constructive-skills pnpm

graphql/query/src/runtime/localhost-fetch.ts
pyramation: should we use our fetch library? what is this for?

pyramation: https://github.com/constructive-io/dev-utils/tree/main/packages/fetch

PR #1283, opened 2026-06-10
fix(presigned-url-plugin): use root connection for storage_module queries

### LINKED ISSUE #1266: presigned-url-plugin uses user connection to query storage_module, tenant users cannot upload
## Problem

Tenant user calling `uploadAppFile` mutation gets error:

```
Error: STORAGE_MODULE_NOT_FOUND
```

But the storage_module record actually exists in the database.

## Execution Flow

```
Tenant user calls uploadAppFile
    ↓
Plugin queries storage_module with tenant user's connection
    ↓
RLS checks org_memberships_sprt
    ↓
Tenant user not in there → returns 0 rows
    ↓
❌ STORAGE_MODULE_NOT_FOUND
```

## Root Cause

**Location**: `graphile/graphile-presigned-url-plugin/src/plugin.ts:277`

```typescript
const allConfigs = await loadAllStorageModules(txClient, databaseId);
//                                             ↑ user connection (with RLS)
```

storage_module's RLS policy checks `org_memberships_sprt` (platform level), but tenant users only exist in `app_memberships_sprt` (tenant level).

**This is correct design**: storage_module is the database owner's system config, tenant users should not directly read it.

**The issue is**: Plugin should not use user connection to read system config.

## Evidence

```sql
-- Platform user can see storage_module
BEGIN;
SET ROLE authenticated;
SELECT set_config('jwt.claims.user_id', '<platform_user_id>', true);
SELECT COUNT(*) FROM metaschema_modules_public.storage_module WHERE database_id = '<db_id>';
-- Returns 1
ROLLBACK;

-- Tenant user cannot see it
BEGIN;
SET ROLE authenticated;
SELECT set_config('jwt.claims.user_id', '<tenant_user_id>', true);
SELECT COUNT(*) FROM metaschema_modules_public.storage_module WHERE database_id = '<db_id>';
-- Returns 0
ROLLBACK;
```

## Why Root Connection is Safe

| Security Point | How It's Checked | Status |
|----------------|------------------|--------|
| databaseId | From `jwt_private.current_database_id()`, not user input | ✅ |
| bucket permission | `getBucketConfig` checks if user can use this bucket | ✅ |
| file RLS | `app_files` table checks `app_memberships_sprt` | ✅ |
| config exposure | User only gets presigned URL, cannot see endpoint/credentials | ✅ |

## Suggested Fix

Modify `presigned-url-plugin` to use root connection for `loadAllStorageModules`:

```typescript
// plugin.ts

// Get rootPgPool from options or context
const rootClient = await getRootPgClient(options);

// Use root to query system config
const allConfigs = await loadAllStorageModules(rootClient, databaseId);

// Other operations (bucket check, file insert) continue using user connection
const bucket = await getBucketConfig(txClient, ...);  // ← Keep RLS
```

**Scope of changes**:
- `graphile-presigned-url-plugin/src/plugin.ts` — use root when querying storage_module
- May need to modify plugin options to pass rootPgPool

## Current Workaround

Manually add RLS policies (test environment only):

```sql
CREATE POLICY auth_select_all ON metaschema_modules_public.storage_module
  FOR SELECT TO authenticated USING (true);
CREATE POLICY auth_select_all ON metaschema_public.schema
  FOR SELECT TO authenticated USING (true);
CREATE POLICY auth_select_all ON metaschema_public.table
  FOR SELECT TO authenticated USING (true);
```

## Impact

- All tenant users cannot use presigned URL upload functionality
- Only platform users can upload files

theothersideofgod:
## Summary
- Tenant users were getting STORAGE_MODULE_NOT_FOUND because RLS on storage_module checks org_memberships_sprt, which only includes platform users
- Fix: Query storage_module with root connection (pgSettings=null), while keeping bucket/file operations under user RLS
- Pattern matches existing codebase (graphile-i18n, PublicKeySignature)

## Security
- databaseId still comes from JWT
- bucket/file RLS unchanged

## Test plan
- [x] Provision database with tenant user
- [x] Upload file as tenant user
- [x] Verify file status changes to 'uploaded'

Fixes #1266

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/graphile/graphile-presigned-url-plugin/src/download-url-field.ts
+++ b/graphile/graphile-presigned-url-plugin/src/download-url-field.ts
@@ -166,16 +166,13 @@ export function createDownloadUrlPlugin(
                               if (!config) return null;
 
                               // Look up the bucket key for scoped S3 resolution
-                              let bucketKey = 'public';
-                              if (bucketId) {
-                                const bucketResult = await rootClient.query({
-                                  text: `SELECT key FROM ${config.bucketsQualifiedName} WHERE id = $1 LIMIT 1`,
-                                  values: [bucketId],
-                                });
-                                if (bucketResult.rows[0]?.key) {
-                                  bucketKey = bucketResult.rows[0].key;
-                                }
-                              }
+                              if (!bucketId) return null;
+                              const bucketResult = await rootClient.query({
+                                text: `SELECT key FROM ${config.bucketsQualifiedName} WHERE id = $1 LIMIT 1`,
+                                values: [bucketId],
+                              });
+                              const bucketKey = bucketResult.rows[0]?.key;
+                              if (!bucketKey) return null;
 
                               return { config, databaseId, bucketKey };
                             });
--- a/graphile/graphile-presigned-url-plugin/src/download-url-field.ts
+++ b/graphile/graphile-presigned-url-plugin/src/download-url-field.ts
@@ -149,33 +149,40 @@ export function createDownloadUrlPlugin(
                       let downloadUrlExpirySeconds = 3600;
                       try {
                         if (withPgClient && pgSettings) {
-                          const resolved = await withPgClient(null, async (pgClient: any) => {
+                          // 1. Get databaseId with user connection (needs JWT claims)
+                          const databaseId = await withPgClient(pgSettings, async (pgClient: any) => {
                             const dbResult = await pgClient.query({
                               text: `SELECT jwt_private.current_database_id() AS id`,
                             });
-                            const databaseId = dbResult.rows[0]?.id;
-                            if (!databaseId) return null;
-                            const allConfigs = await loadAllStorageModules(pgClient, databaseId);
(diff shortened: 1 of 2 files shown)


theothersideofgod:
## Fixed: Remove hardcoded 'public' bucket key

@pyramation Thanks for the review!

**Problem**: `let bucketKey = 'public'` as fallback would store files in wrong S3 path

**Fix** (commit `dafc1b72c`):
- Require `bucketId` and lookup `key` from database
- Return null if not found, falling back to global config
- No more default `'public'`

```diff
- let bucketKey = 'public';
- if (bucketId) {
-   const bucketResult = await rootClient.query(...);
-   if (bucketResult.rows[0]?.key) {
-     bucketKey = bucketResult.rows[0].key;
-   }
- }
+ if (!bucketId) return null;
+ const bucketResult = await rootClient.query(...);
+ const bucketKey = bucketResult.rows[0]?.key;
+ if (!bucketKey) return null;
```

PR #1308, opened 2026-06-19
Fix GraphQL private header routing priority mismatch

Zetazzz:
## Summary

Align private GraphQL API routing priority between `getSvcKey()` and `determineMode()`.

Previously, requests containing both `X-Api-Name` and `X-Schemata` could be cached under an `api:<db>:<apiName>` service key while being resolved through the `schemata-header` path. That could store an admin-style schema-header API structure under a named API cache key.

This PR makes both paths use the same private header priority:

`X-Api-Name > X-Schemata > X-Meta-Schema`

## Changes

- Added a shared private header mode helper used by both service-key generation and mode selection.
- Kept existing `X-Api-Name` service-key semantics unchanged.
- Added regression tests for mixed `X-Api-Name` / `X-Schemata` requests.

## Validation

- `pnpm --filter @constructive-io/graphql-server exec jest src/middleware/__tests__/api.test.ts --runInBand`
- `pnpm --filter @constructive-io/graphql-server build`
- `pnpm --filter @constructive-io/graphql-server exec jest --runInBand`

### DIFF
--- a/graphql/server/src/middleware/__tests__/api.test.ts
+++ b/graphql/server/src/middleware/__tests__/api.test.ts
@@ -1,0 +1,117 @@
+jest.mock('pg-cache', () => ({
+  getPgPool: jest.fn(),
+}));
+
+jest.mock('@constructive-io/express-context', () => ({
+  createDefaultRegistry: jest.fn(() => ({
+    resolve: jest.fn().mockResolvedValue(undefined),
+  })),
+}));
+
+import type { Request } from 'express';
+import type { Pool } from 'pg';
+import { svcCache } from '@pgpmjs/server-utils';
+import { getPgPool } from 'pg-cache';
+
+import type { ApiOptions } from '../../types';
+import { getApiConfig, getSvcKey } from '../api';
+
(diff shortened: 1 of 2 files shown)


Zetazzz:
in short, this's a correctness for a corner case in the private mode:

in getting svckey, the priority was:
if (req.get('X-Api-Name')) return `api:...`;
if (req.get('X-Schemata')) return `schemata:...`;
if (req.get('X-Meta-Schema')) return `metaschema:...`;

  X-Api-Name > X-Schemata > X-Meta-Schema

and in getting the value(its mode), the priority was:
    if (headers.schemata) return 'schemata-header';
    if (headers.apiName) return 'api-name-header';
    if (headers.metaSchema) return 'meta-schema-header';

  X-Schemata > X-Api-Name > X-Meta-Schema

so there'll be a possible missmatch when there're both 'X-Api-Name' and 'X-Schemata':
the key will be 'api:....'
and the value will be 'schemata-header'

The PR is a small correctness fix with a minor refactor: it centralizes the private header priority so key generation and mode selection cannot diverge again.

PR #1325, opened 2026-07-02
(DO NOT MERGE: superseded by stacked PRs) (WIP) proof of concept: feat(server): heap-safe instance cache + opt-in blueprint pooling (one instance per schema-shape)

yyyyaaa:
## Summary

Makes the multi-tenant PostGraphile v5 server survive and scale on small heaps. Two layers:

1. **Availability hardening (always on):** heap-aware instance cache with safe eviction — the server degrades gracefully under tenant churn instead of OOMing.
2. **Blueprint pooling (opt-in, `GRAPHILE_BLUEPRINT_POOLING=1`):** one shared PostGraphile instance per *schema-shape*, routed per request via `search_path` — collapsing N same-shape tenants from N×~0.5GB to one instance.

Also fixes three latent multi-tenant bugs found during the audit (one live RLS bypass).

## Problem

Each query-serving PostGraphile instance retains ~0.5GB of heap (measured; ~51% strings, ~30% plan closures). The cache was keyed per `svc_key` (per tenant×API) with `max: 50`, TTL 1 year — a steady state of ~24GB for a fleet that runs in 2GB containers. Memory grew linearly with tenants and the process OOM'd long before "thousands of tenants."

### 1. Cache + eviction hardening (`graphile-cache`, `graphql/server`)
- Heap-aware default cap: `clamp(⌊heap×0.5 / 512MB⌋, 3, 50)` instead of a fixed 50 (`GRAPHILE_CACHE_MAX`, `GRAPHILE_CACHE_INSTANCE_HEAP_BYTES` override/tune); resolved cap logged at startup.
- Entry-identity disposal guard (fixes a same-key disposal race), pool-coupling by `entry.dbname` (the old substring match never fired).
- **Eviction drain:** in-flight requests are refcounted (`invokeEntryHandler`); `disposeEntry` waits for them (bounded by `GRAPHILE_CACHE_DRAIN_TIMEOUT_MS`, default 30s) before `pgl.release()` — eviction can no longer tear a schema down mid-request.
- **Build admission control:** cross-key builds serialize through a semaphore (`GRAPHILE_BUILD_CONCURRENCY`, default 1) and evict the LRU instance *before* building, so the build's transient peak lands on freed headroom.
- Prod idle TTL 1 year → 6h; GraphiQL (ruru) off outside development (`GRAPHILE_GRAPHIQL=true` to force).

### 2. Multi-tenant bug fixes
- **`graphile-i18n`: live RLS bypass** — the `localeStrings` query ran `withPgClient(null, …)` (no role, no claims). Now threads the request's `pgSettings`. Also resets the type cache per build (module-singleton leak).
- **meta-schema:** `cachedTablesMeta` was a module global — concurrent builds could serve each other's `_meta`. Now keyed per build via WeakMap (build objects are frozen; the flat global remains only for single-build codegen consumers, documented).
- **`graphile-llm` agent-discovery:** config cache was keyed by dbname with a `LIMIT 1`, no tenant filter — cross-tenant bleed in shared-DB topologies. Now filtered and keyed by `database_id`.
- **`graphile-presigned-url`:** storage-module resolution matched build-time physical schema names; now matches logical names (hash prefix stripped) so it works under pooling and across re-hashed schemas.

### 3. Blueprint pooling (opt-in)
- **Key:** `bp:sha256({sorted logical schemas, shape fingerprint, database settings flags, api name, mode, dbname})`. The shape fingerprint hashes the catalog's `[logical schema, relname]` pairs, so tenants that drifted (e.g. a half-provisioned tenant) automatically get their own instance. `dbname` is included so same-shape tenants in different physical databases never share a pool.
- **Mechanism:** shared instances build with the stock `gather: { pgIdentifiers: 'unqualified' }` (search_path-relative SQL — GraphQL SDL is byte-identical to qualified builds, sha256-verified) plus `schema: { constructiveUnqualified: true }` so Constructive plugins (search chunk refs + BM25 index name, llm RAG chunk query, i18n localeStrings) emit search_path-relative SQL for tenant data. Control-plane (`metaschema_*`, `services_public`) references stay fully qualified by design.
- **Routing:** `grafast.context` reads roles from `req.api` (de-closured in all modes) and, on pooled instances only, sets `pgSettings.search_path` = requesting tenant's physical schemas + `public` last (shared domains/extensions — SECURITY DEFINER functions like `sign_in` depend on it).
- **Safety fallbacks to today's per-tenant instances:** realtime-enabled APIs; empty schema lists; unqualified relation-name collisions within the schema set (e.g. the intentional `identity_providers` table/view shadow — detected by catalog probe, logged, per-tenant instance used); failed probes (not memoized — re-probed next request).
- **Invalidation (v1):** any tenant `schema:update` flushes all pooled instances + cached decisions (rebuilds are ~1–2s for tenant APIs); the manual `/flush` route does the same.
- Flag **off** is verified behavior-identical (existing middleware suites unchanged; plugin emissions byte-identical).

## Verification (isolated rig: full constructive-db schema + 8 seeded marketplace tenants as hashed schemas, server at 2GB heap)

- **Gate:** SDL qualified vs unqualified — byte-identical (same sha256). Zero-bleed spike at the engine level — one unqualified instance served other tenants purely via search_path, canaries clean.
- **Flag OFF regression:** 40/40 requests, one instance per host, no pooling lines; separately 360/360 requests across two eviction soaks with 251+ pre-build evictions, 0 drain timeouts, 0 5xx.
- **Flag ON:** 7 same-shape tenants → **one** shared instance (6 builds → 1 for the same hosts; 56 requests served by one build); the half-provisioned tenant automatically split to its own instance by fingerprint. **Zero-bleed via HTTP-authenticated canary queries** through the shared instance (10 interleaved rounds; cross-token control correctly unauthenticated). RSS for the same traffic: 1475MB (per-tenant) → **711MB** (pooled). Collision fallback observed live (warn + per-tenant instance). Pooled login verified end-to-end (signIn → bearer → tenant data).
- **Adversarial review:** 3 dimensions (correctness / tenant isolation / perf-ops), 19 findings adjudicated by refute-by-default verifiers → 9 confirmed → 6 fixed in this PR (search_path `public` drop, dbname in key, transient-probe memoization, `/flush` bp: gap, double catalog scan, `_meta` physical-name leak), 3 documented below.

## Env vars

| Var | Default | Purpose |
|---|---|---|
| `GRAPHILE_BLUEPRINT_POOLING` | off | Opt-in instance sharing per schema-shape |
| `GRAPHILE_CACHE_MAX` | heap-aware | Max cached instances |
| `GRAPHILE_CACHE_INSTANCE_HEAP_BYTES` | 512MB | Per-instance estimate for the heap-aware cap |
| `GRAPHILE_CACHE_TTL_MS` | 6h prod / 5m dev | Idle TTL |
| `GRAPHILE_CACHE_DRAIN_TIMEOUT_MS` | 30s | Max wait for in-flight requests before release |
| `GRAPHILE_BUILD_CONCURRENCY` | 1 | Concurrent schema builds |
| `GRAPHILE_GRAPHIQL` | off in prod | Force-enable GraphiQL |

## Known limitations / follow-ups

- **Fingerprint granularity:** relation names only (not columns/functions/enum labels). Same-relname column drift between pooled tenants is not detected by the key; exposure is bounded by flush-all-on-`schema:update`. Follow-up: extend the fingerprint to attributes/procs.
- **v1 flush semantics** are coarse (any tenant migration flushes all pooled instances). Cheap rebuilds make this acceptable; follow-up: blueprint-membership tracking for targeted eviction.
- **Build semaphore** has no wall-clock timeout — a pathologically slow build delays queued builds (requests on cached instances are unaffected). Follow-up: per-build timeout + 503.
- Headroom accounting counts cached instances only, not still-draining ones (bounded by the 30s drain timeout).
- Realtime APIs are excluded from pooling (per-instance LISTEN topology); pooling them needs the channel scheme rework.
- The `/flush` route's missing auth is pre-existing (`TODO` in code).

## Rollout

1. Ship with flag off (behavior-identical; hardening + bug fixes active).
2. Staging: `GRAPHILE_BLUEPRINT_POOLING=1`, watch `[pooling]` logs (attach vs build), instance counts, RSS.
3. Production canary on public API pods; auth/admin endpoints pool automatically where collision-free.


graphile/graphile-i18n/src/plugin.ts
```diff
@@ -238,10 +239,22 @@ export function createI18nPlugin(options: I18nPluginOptions = {}): GraphileConfi
...
+          // (schema.constructiveUnqualified), emit search_path-relative table
+          // references so the per-request search_path resolves the tenant schema.
+          // Default (flag absent): fully schema-qualified, byte-identical output.
+          const constructiveUnqualified = !!(build.options as any).constructiveUnqualified;
+          const baseTableRef = constructiveUnqualified
+            ? `"${baseTable}"`
+            : `"${schemaName}"."${baseTable}"`;
+          const translationTableRef = constructiveUnqualified
+            ? `"${translationTable}"`
+            : `"${schemaName}"."${translationTable}"`;
```
pyramation: let's aim to use this where we can instead of just quoting stuff w/o checking:

https://www.npmjs.com/package/@pgsql/quotes

```ts
QuoteUtils.quoteQualifiedIdentifier('public', 'my_table');
```

graphile/graphile-llm/src/plugins/agent-discovery-plugin.ts
```diff
@@ -49,26 +52,39 @@ const DISCOVERY_SQL = `
     acm.task_table_name
   FROM metaschema_modules_public.agent_chat_module acm
   JOIN metaschema_public.schema s ON s.id = acm.schema_id
+  WHERE s.database_id = $1
```
pyramation, line 55: this is great, not sure if we can separate a few smaller PRs, would be great :)

pyramation: if not, I also get it, we can take a look together then.

graphile/graphile-search/src/adapters/bm25.ts
```diff
@@ -169,26 +169,37 @@ export function createBm25Adapter(
...
       if (filterValue == null) return null;
 
       const { query, threshold, includeChunks } = filterValue;
       if (!query || typeof query !== 'string' || query.trim().length === 0) return null;
 
+      // When the instance is built for blueprint pooling
+      // (schema.constructiveUnqualified), emit search_path-relative references for
+      // tenant-data tables/indexes so the per-request search_path resolves the
+      // tenant schema. Default (flag absent): fully schema-qualified, byte-identical.
+      const constructiveUnqualified = !!((build as any)?.options?.constructiveUnqualified);
```
pyramation: wait, this means using unqualified schemas?

PR #1330, opened 2026-07-06
fix(graphile): tenant-isolation fixes for i18n/llm/search/meta-schema + @pgsql/quotes adoption (1/6)

yyyyaaa:
## Summary

First PR of the 6-PR stack that splits #1325 into incrementally mergeable pieces (per review feedback there). This one carries the four **standalone correctness/tenant-isolation bugs** the #1325 audit found in plugins — every one of them is live behavior today and none depends on caching or pooling — plus the requested **@pgsql/quotes adoption**.

## Fixes

- **graphile-i18n — RLS bypass.** The `localeStrings` translations query ran through `withPgClient(null, …)`: no role, no JWT claims, so RLS on the base + translation tables was not applied for that read. It now passes the request's `pgSettings`. Also: the locale `GraphQLObjectType` cache is reset per build (`init` hook) so a rebuilt schema can't reuse type objects from a previous build.
- **graphile-llm (agent-discovery) — cross-database config leak.** The discovery query had no database filter: `LIMIT 1` over a control-plane table holding *every* database's rows, so a process serving multiple databases could return another database's agent config. Now filtered by the requester's `jwt.claims.database_id`, failing closed when absent, with the 60s cache keyed by database id.
- **graphile-settings (meta-schema) — concurrent-build `_meta` bleed.** Collected table metadata lived in a module-global; concurrent schema builds interleave `init` hooks, so one build's `_meta` could serve another build's tables. Now keyed by the `build` object (WeakMap). The flat mirror stays for out-of-process codegen consumers (single-build, last-write-wins is safe there).
- **graphile-llm (rag) / graphile-search (bm25) — qualification hygiene.** Chunk-table and bm25-index references stay schema-qualified; SQL builders exported for unit testing.

## @pgsql/quotes adoption

Addresses the review comment on `graphile-i18n/src/plugin.ts` in #1325. Hand-rolled `"${schema}"."${table}"` templates in i18n, rag, and bm25 are replaced with `QuoteUtils.quoteIdentifier` / `quoteQualifiedIdentifier`.

Two properties worth knowing:

- The library ports PostgreSQL's `quote_ident`: identifiers come out **bare when lexically safe** and double-quoted (with `"` doubled) when required — strictly more correct than always-quote templates for case-sensitive names, reserved keywords, and embedded quotes. Test assertions were updated for the bare snake_case forms.
- **Load-bearing and now pinned by tests:** hashed multi-tenant schema names contain `-`, so they are *always* emitted double-quoted. The pooling rewrite seam (later in this stack) matches double-quoted schema tokens in SQL text, so tenant-schema references in plugin SQL must never be emitted bare. New test cases assert the dashed-schema quoting in both rag and bm25.

Where it is deliberately **not** used: the rewrite seam itself (PR 4 of the stack) must always emit quoted identifiers when substituting tokens, so `quote_ident`-style bare output is wrong there by construction.

`@pgsql/quotes@17.1.0` is already in the workspace dependency graph (graphile-settings, presigned-url, realtime, pgpm/cli) — zero new install weight; the three plugin package.jsons just declare it.

## Commits

1. The four fixes (verbatim from #1325).
2. `chore(lockfile)` — pnpm 10.30 re-normalization with **no dependency changes**, split out so the real deltas stay readable (~10k mechanical lines; the same churn is embedded in #1325's lockfile diff).
3. The quotes adoption (+137/−31 with a 44-line lockfile delta).

## Testing

`graphile-i18n` 5/5 · `graphile-llm` 58/58 non-Ollama (11 env-dependent Ollama integration tests need the CI service container) incl. 11 new agent-discovery/rag tests · `graphile-search` 78/78 incl. new quoting tests · `graphile-settings` 158/158.

## Stack

| # | branch | contents |
|---|--------|----------|
| **1 (this)** | #1330 `feat/scale-s1-plugin-fixes` | plugin correctness fixes + @pgsql/quotes |
| 2 | #1331 `feat/scale-s2-cache-hardening` | heap-safe cache, memory governor, build admission |
| 3 | #1332 `feat/scale-s3-introspection-filter` | namespace-scoped introspection (87× per-instance memory) |
| 4 | #1333 `feat/scale-s4-pooling-core` | blueprint pooling core — rewrite seam (inert modules) |
| 5 | #1334 `feat/scale-s5-pooling-integration` | opt-in pooling integration (`GRAPHILE_BLUEPRINT_POOLING`) |
| 6 | #1335 `feat/scale-s6-perf-tooling` | cperf harness + validation rig + capacity docs |

Each PR targets the previous branch; merge top-to-bottom (GitHub retargets children automatically when a base merges). Extracted from #1325, which stays open as the validated end-to-end reference — all numbers cited in this stack were measured there on a production-shaped rig (48 tenant DBs, 61k `pg_class` rows).

### DIFF
--- a/agentic/agentic-server/package.json
+++ b/agentic/agentic-server/package.json
@@ -1,7 +1,7 @@
 {
   "name": "agentic-server",
   "version": "0.8.0",
-  "description": "Standalone Express LLM service — agent threads, chat streaming, billing metering, and inference logging via express-context",
+  "description": "Standalone Express LLM service - agent threads, chat streaming, billing metering, and inference logging via express-context",
   "main": "index.js",
   "module": "esm/index.js",
   "types": "index.d.ts",
--- a/graphile/graphile-bucket-provisioner-plugin/package.json
+++ b/graphile/graphile-bucket-provisioner-plugin/package.json
@@ -1,7 +1,7 @@
 {
   "name": "graphile-bucket-provisioner-plugin",
   "version": "0.13.0",
-  "description": "Bucket provisioning plugin for PostGraphile v5 — auto-provisions S3 buckets on bucket table mutations",
+  "description": "Bucket provisioning plugin for PostGraphile v5 - auto-provisions S3 buckets on bucket table mutations",
   "author": "Constructive <developers@constructive.io>",
   "homepage": "https://github.com/constructive-io/constructive",
   "license": "MIT",
--- a/graphile/graphile-i18n/package.json
+++ b/graphile/graphile-i18n/package.json
@@ -1,7 +1,7 @@
 {
   "name": "graphile-i18n",
   "version": "1.4.1",
-  "description": "PostGraphile v5 i18n plugin — language-aware fields from @i18n translation tables with Accept-Language negotiation and fallback chains",
+  "description": "PostGraphile v5 i18n plugin \u2014 language-aware fields from @i18n translation tables with Accept-Language negotiation and fallback chains",
   "author": "Constructive <developers@constructive.io>",
   "homepage": "https://github.com/constructive-io/constructive",
   "license": "MIT",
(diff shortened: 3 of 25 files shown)


yyyyaaa:

PR #1331, opened 2026-07-06
feat(graphile-cache,server): heap-safe instance cache, memory governor, build admission control (2/6)

yyyyaaa:
## Summary

PR 2/6 of the #1325 split. Makes PostGraphile instance residency **heap-aware** and turns every memory degradation mode into a `503 + Retry-After` instead of a process abort. No pooling here — this hardens the existing per-service cache and is independently shippable; it's the "server stays alive" layer. Every mechanism traces to a specific measured crash in the validation program (`scripts/scale-validate/VALIDATION-REPORT.md`, lands with PR 6).

**Why:** on a production-shaped catalog (61k `pg_class` rows), one cached instance retains up to **~1.3GB**. The previous cache admitted a fixed **50 entries with a 1-year TTL** and disposed entries that still had requests in flight — on catalogs this size that configuration is the OOM, not a cache.

## graphile-cache

- **Heap-aware capacity** replaces `max=50`: `R = clamp(min(⌊(H−B)/I⌋, ⌊(H−B−T)/I⌋+1), 1, 50)` with B=256MB server baseline, T=768MB build transient, I = per-instance budget (`GRAPHILE_CACHE_INSTANCE_HEAP_BYTES`).
- **In-flight refcounting + eviction drain** (`invokeEntryHandler`): disposal waits (bounded) for the entry's live requests instead of closing pools under them.
- **Memory governor**: background eviction at 85% pressure, build refusal at 92%. Pressure is `used / (used + total_available_size)` — the earlier `heap_size_limit` denominator was **unreachable in practice** (measured death at ratio 0.69 on a 512MB heap, long before a 0.92 trigger could fire).
- Entries carry their `dbname` so pool disposal evicts exactly the entries it strands.

## graphql/server

- **Process-wide build semaphore** (`GRAPHILE_BUILD_CONCURRENCY`, default 1): the single-flight map only dedupes builds of the *same* key, so N cold-starting tenants used to stack N several-hundred-MB build transients. Plus **evict-before-build**, **drain-wait**, and a **late pressure re-check** with elevated-pressure spacing (`GRAPHILE_BUILD_PRESSURE_SPACING_MS`) so build starts can't outrun GC.
- **Request build-wait timeout** (`GRAPHILE_BUILD_TIMEOUT_MS`, default 180s) via `raceBuildAgainstTimeout`, which **always clears its timer**. This is load-bearing: an armed `Promise.race` timeout pins the freshly built instance through the timer's reaction chain for the full 180s. Under eviction churn that stacked ~2 builds/s × retained instance × 180s and OOM'd the process — root-caused by heap-snapshot retainer chains, and `.unref()` demonstrably does *not* help (it removes event-loop keepalive, not GC reachability).
- **Build settles own the cache**: the single-flight slot is cleared when the *build* settles, not when the first requester stops waiting — timed-out waiters leave coalescing intact for followers.
- **Uniform 503 taxonomy**: `SERVICE_OVERLOADED` (pressure refusal — at the request gate, the build owner path, and re-coalesced waiters alike), `BUILD_TIMEOUT` (wait expired; build continues in background and fills the cache for retries), `SERVICE_ROTATING` (mid-disposal hit). Never a generic 500 for a governed condition.
- **Connection-class error guard**: a PostgreSQL restart degrades requests instead of killing the server process.
- **Observability**: `/metrics` endpoint (`GRAPHILE_METRICS_ENDPOINT=1`, loopback-gated), JSON-line sampler (`GRAPHILE_DEBUG_METRICS`) with GC pause tracking and build/cache counters. GraphiQL (ruru) assets are now dev-only (`GRAPHILE_GRAPHIQL=true` to force) — they were per-instance overhead and an unnecessary prod surface.

## Validation (from #1325)

- Eviction-churn storm at ~23 builds/s for 240s+: bounded heap sawtooth, zero OOM (pre-fix: 2GB heap dead in 65s).
- Capacity rotation 1444/1444 OK; 47-tenant 30min soak headline in PR 5.

## Testing

`graphile-cache` 24/24 (counters, disposal/cap, governor) · `graphql/server` 119/119 including new `build-timeout` (timer-count-zero assertions) and `graphile-recoalesce` (503 contract, proven 500 pre-fix) suites, plus all pre-existing middleware suites.

## Stack

Stacked on #1330 (`feat/scale-s1-plugin-fixes`); merge after it. Full stack map in #1330. Extracted from #1325 (kept open as the validated reference).

### DIFF
--- a/graphile/graphile-cache/src/__tests__/counters.test.ts
+++ b/graphile/graphile-cache/src/__tests__/counters.test.ts
@@ -1,0 +1,120 @@
+import {
+  cacheCounters,
+  clearMatchingEntries,
+  getCacheConfig,
+  getCacheCounters,
+  graphileCache,
+  type GraphileCacheEntry
+} from '../graphile-cache';
+
+/**
+ * Counter instrumentation regression tests. The counters are process-global and
+ * mutable, so every assertion is a DELTA against a snapshot captured immediately
+ * before the action — robust to test ordering and to fire-and-forget disposals from
+ * other suites. The eviction and disposal increments happen synchronously inside
+ * graphileCache.delete() (the dispose callback and the pre-await head of disposeEntry),
+ * so no tick is needed except for the async drain-timeout path.
+ */
+
+const tick = (ms = 60) => new Promise((resolve) => setTimeout(resolve, ms));
+
+const makeEntry = (releaseDelayMs = 0): GraphileCacheEntry => {
+  const release = jest.fn(
+    () => new Promise<void>((resolve) => setTimeout(resolve, releaseDelayMs))
+  );
+  return {
+    pgl: { release } as unknown as GraphileCacheEntry['pgl'],
+    serv: {} as GraphileCacheEntry['serv'],
+    handler: {} as GraphileCacheEntry['handler'],
+    // Never .listen()ed in production; close() invokes its callback synchronously here.
+    httpServer: { close: (cb: () => void) => cb() } as unknown as GraphileCacheEntry['httpServer'],
+    cacheKey: 'mock',
+    createdAt: Date.now()
+  };
+};
+
+describe('cacheCounters increment on eviction/disposal', () => {
+  afterEach(async () => {
+    graphileCache.clear();
+    await tick();
+  });
+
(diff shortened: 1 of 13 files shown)


yyyyaaa:

PR #1332, opened 2026-07-06
feat(server): namespace-scoped introspection filter — 87× per-instance memory (3/6)

yyyyaaa:
## Summary

PR 3/6 of the #1325 split. Scopes each instance's catalog introspection to the namespaces it actually serves, opt-in via `GRAPHILE_INTROSPECTION_FILTER=1`. This turned out to be **the** memory wall — and it applies to **dedicated (non-pooled) instances**, so it delivers most of the per-instance win even if pooling (PRs 4–5) never merges.

Measured on the #1325 rig (48 tenant DBs, 61k `pg_class` rows in one database):

| metric | before | after | |
|---|---|---|---|
| retained heap per instance | 1,305.6MB | **14.7MB** | 87× |
| cold build | 7.2s | **0.42s** | 17× |
| PG-side introspection spike | +1,283MiB | **+17MiB** | |

## Mechanism (`introspection-filter.ts`)

- The stock `pg-introspection` query gates namespaces with a fixed textual predicate that appears **exactly four times**. Pre-parse, each occurrence is swapped for `nspname = any (array[…]::text[])` over the keep set.
- Keep set = served schemas ∪ `public` ∪ a bounded closure over cross-schema references (FK targets, attribute/procedure type owners), discovered with **qualified** catalog queries (`MAX_CLOSURE_ROUNDS=5`) and memoized per pool.
- **Fail-open on gate mismatch**: if a pg-introspection upgrade changes the text, introspection runs unfiltered (correctness over savings) — and the pinning test fails loudly in CI instead of the filter silently disabling in production. **Fail-closed on discovery errors.**
- `pg-introspection@1.0.1` is pinned as a devDependency purely so the test suite can assert the installed package still carries the exact gate text.

## Wiring

~10 lines in `buildPreset`: when the flag is on and the instance has a concrete schema list, the pgService pool gets a thin interceptor that touches only the introspection statement; every other query passes through byte-identical. Flag off ⇒ pool selection identical to today. Metrics sampler gains swap/discovery counters.

## Testing

`graphql/server` 132/132 (14 suites) — includes the pinning test, gate-mismatch fail-open, literal escaping (quote doubling, `pg_*`/`information_schema` drops), closure-loop memoization, and the dedicated-wrapper path. (The filter-through-rewriting-pool composition tests land with PR 4, where that seam exists.)

## Stack

Stacked on #1331 (`feat/scale-s2-cache-hardening`); merge after it. Full stack map in #1330. Extracted from #1325 (commit 7615a8c00; kept open as the validated reference).

### DIFF
--- a/graphql/server/package.json
+++ b/graphql/server/package.json
@@ -88,6 +88,7 @@
     "graphile-test": "workspace:*",
     "makage": "^0.3.0",
     "nodemon": "^3.1.14",
+    "pg-introspection": "1.0.1",
     "ts-node": "^10.9.2"
   }
 }
--- a/graphql/server/src/diagnostics/__tests__/metrics-sampler.test.ts
+++ b/graphql/server/src/diagnostics/__tests__/metrics-sampler.test.ts
@@ -47,6 +47,8 @@ describe('collectMetricsSample', () => {
     expect(typeof sample.counters.evictions.lru).toBe('number');
     expect(typeof sample.counters.builds).toBe('number');
     expect(typeof sample.counters.buildQueueDepth).toBe('number');
+    expect(typeof sample.counters.introspectionFilter.swaps).toBe('number');
+    expect(typeof sample.counters.introspectionFilter.discoveries).toBe('number');
     expect(typeof sample.gc).toBe('object');
 
     // Round-trips through JSON without loss.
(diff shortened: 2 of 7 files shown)


yyyyaaa:

PR #1630, opened 2026-08-03
feat(pgpm): ledger-aware diff — workspace sides, pgpm_migrate classification, ledger backfill

pyramation:
## Summary

Ledger-aware `pgpm diff`: relate a plan-bearing side to a live database's `pgpm_migrate` ledger, recognize regenerated/renamed/reordered plans as already-satisfied via semantic identity, and emit a log-only backfill so a subsequent `pgpm deploy` executes only the genuine delta.

```
workspace + ledger → classify (name/hash) × cover (semantic identity)
                   → backfill (satisfied ∧ unrecorded, log-only)
                   → deploy executes only the true delta
```

### Layering (post-review refactor: adapters live in core)

- **`@pgpmjs/diff` (pure):** `DiffSide` normalization, `classifyAgainstLedger` (deployed-identical / drifted / pending / orphaned / out-of-order), `selectBackfillEntries` (pending × satisfied → backfill; inert/partial reported as unprovable), `emitLedgerBackfill` (idempotent `CALL pgpm_migrate.deploy(..., p_log_only => TRUE)` script), `resolveDiffSideKind` (the canonical workspace/module/sql/db classifier).
- **`@pgpmjs/transform`:** `coverChanges` — per-change semantic coverage (`satisfied|unsatisfied|partial|inert`) keyed by object identity, so plan regeneration washes out.
- **`@pgpmjs/core` (I/O adapters, new `src/diff/` + `src/dump/`):**
  - `loadDiffSide(spec, cwd)` — any side: db (via `pg_dump`), workspace (flattened with `PgpmPackage.resolveWorkspaceExtensionDependencies`, the same resolver deploy uses), module, .sql.
  - `planChangeRefs` (content + AST hash candidates), `readLedgerSnapshot` (`PgpmMigrate.readDeployedState`), `buildLedgerReport` (classification + coverage + backfill).
  - `pgDump(options)` — typed pg_dump helper (schemaOnly/dataOnly, inserts/columnInserts, schema/table include/exclude, file or stdout, `command`/`PGPM_PG_DUMP` override for a version-matched containerized binary); used by both `pgpm dump` and `pgpm diff`. `dumpSchemaForComparison` excludes `LEDGER_SCHEMAS` (`pgpm_migrate`, `sqitch`) — ledger metadata is read separately, not diffed as schema.
  - `resolveDatabaseSpec`/`isDatabaseSpec` — `db:<name>` and DSN specs resolved through `pg-env` layering.
- **`pgpm` CLI:** `diff.ts` is arg-parsing + printing + orchestration only; hand-rolled `isWorkspaceDir`/DSN parse/pg_dump spawns deleted.

### CLI

```
pgpm diff db:prod ./workspace-v2 --ledger db:prod --emit-ledger ./backfill.sql \
         --emit-migration ./out --pkg upgrade
```

`--ledger`/`--emit-ledger` are opt-in; plain `pgpm diff A B` behavior is unchanged.

### Flagship e2e

`pgpm/cli/__tests__/diff-ledger-e2e.test.ts`: deploys a v1 generated-style workspace to "prod", regenerates v2 (all plans renamed/reordered/regrouped + one new table), classifies against the ledger, applies the backfill, then deploys v2 — only the new table executes, and the final catalog equals a fresh v2 deploy. CI runs `pg_dump` from the `postgres-plus:18` service container via `PGPM_PG_DUMP` (no client install).

Link to Devin session: https://app.devin.ai/sessions/34f64615befa49d7b807f3467a07e726
Requested by: @pyramation


pgpm/cli/src/commands/diff.ts
```diff
@@ -90,23 +109,52 @@ Examples:
...
 type NamingStyle = (typeof NAMING_STYLES)[number];
 
-/** pg_dump a side's schema: `db:<name>` uses PG* env; DSNs pass through. */
+/**
+ * The pg_dump command to run. Defaults to the `pg_dump` on PATH, but honours
+ * PGPM_PG_DUMP — a whitespace-separated command such as
+ * `docker exec <container> pg_dump` — so callers can run a version-matched
+ * pg_dump from a Postgres container instead of installing client tools.
+ */
+const resolvePgDump = (): { cmd: string; prefixArgs: string[] } => {
```
pyramation: how is this used?

pgpm/cli/src/commands/diff.ts
```diff
@@ -90,23 +109,52 @@ Examples:
...
+  const override = (process.env.PGPM_PG_DUMP ?? '').trim();
```
pyramation: why env var for this? is this for CI? why not using 12factor-env, etc? is that because this is ephemeral for a one off job?

pgpm/cli/src/commands/diff.ts
```diff
@@ -90,23 +109,52 @@ Examples:
...
+};
+
+/**
+ * pg_dump a side's schema: `db:<name>` uses PG* env; DSNs pass through.
+ * Migration metadata schemas (pgpm_migrate, sqitch) are excluded — they are
+ * the ledger, not the schema under comparison.
+ */
 const dumpDatabase = async (spec: string): Promise<string> => {
-  const args = ['--schema-only', '--no-owner'];
+  const args = ['--schema-only', '--no-owner', '--exclude-schema=pgpm_migrate', '--exclude-schema=sqitch'];
```
pyramation: Interesting, so if you exclude these schemas, then that means you're not migrating them in? Just because they're normally not migrated? How do you then read them?

pyramation: Also, another thing is that we may need more options for this too, like inserts only and things like that. While we're at it, should we make this a little bit more robust?

pgpm/cli/src/commands/diff.ts
```diff
@@ -136,7 +184,29 @@ const loadSide = async (spec: string, cwd: string): Promise<DiffSide> => {
...
     };
   }
-  return loadDiffSideFromDisk(path.resolve(cwd, spec));
+  const resolved = path.resolve(cwd, spec);
+  if (isWorkspaceDir(resolved)) return loadWorkspaceSide(resolved);
+  return loadDiffSideFromDisk(resolved);
+};
+
+/** A directory that is a workspace root rather than a module root. */
+const isWorkspaceDir = (dir: string): boolean =>
```
pyramation: I am nearly certain we must already have something like this. If we don't, we should find out which area it should be in and put it there. I think we should do a little bit of research on this because I don't think we need to create this from scratch.

pgpm/cli/src/commands/diff.ts
```diff
@@ -136,7 +184,29 @@ const loadSide = async (spec: string, cwd: string): Promise<DiffSide> => {
...
+  fs.existsSync(dir) &&
+  fs.statSync(dir).isDirectory() &&
+  !fs.existsSync(path.join(dir, 'pgpm.plan')) &&
+  ['pgpm.json', 'pgpm.config.js'].some(f => fs.existsSync(path.join(dir, f)));
+
+/** PgConfig for a `db:<name>` spec or a postgres:// connection string. */
+const ledgerConfig = (spec: string): PgConfig => {
+  if (spec.startsWith('db:')) return getPgEnvOptions({ database: spec.slice(3) });
```
pyramation: Which files do we have to do that? Who inserts the string, and how do we know it's going to have a DB and colon in it?

pgpm/cli/src/utils/diff-sides.ts
```diff
@@ -0,0 +1,105 @@
...
+import * as path from 'path';
+import type { PgConfig } from 'pg-env';
+
+/**
+ * Load every local module of a pgpm workspace, in cross-module dependency
+ * order, as `ModuleSource`s. External extensions (installed Postgres
+ * extensions, not local modules) are skipped: they are not part of the
+ * workspace's authored schema.
+ */
+export const loadWorkspaceModules = async (
```
pyramation: Why did this need to be built? Do we not have someone that does this? The name of this function is so generic that I can't believe we wouldn't have something like this already written.

pgpm/cli/src/utils/diff-sides.ts
```diff
@@ -0,0 +1,105 @@
...
+  };
+};
+
+/**
+ * Candidate hashes for every plan entry of a set of workspace modules: the
+ * raw-content hash (what `pgpm deploy` records by default) and the AST hash
+ * (what `DEPLOYMENT_HASH_METHOD=ast` records — formatting- and comment-proof).
+ * A ledger row matches when it recorded either.
+ */
+export const planChangeRefs = async (
```
pyramation: Nothing like this art existed? 

Generally, I feel like I'm seeing just way too much code in CLI. 

Typically, we would write a lot of this kind of thing in the core, so I'm kinda tempted to close this pull request as a result. Can you just double-check the core and make sure there are not all these things that you're re-implementing in the CLI already existing in the core?

pgpm/diff/src/ledger.ts
```diff
@@ -0,0 +1,156 @@
+/**
+ * Ledger-aware classification and backfill for `pgpm diff`.
+ *
+ * The `pgpm_migrate` ledger is a cursor into a workspace's deployment
+ * history: which plan changes a database has executed, with what script
+ * hashes, in what order. These pure helpers relate a workspace's plan to a
+ * ledger snapshot (classification) and project a set of already-satisfied
+ * changes into idempotent ledger writes (backfill) so a database can adopt a
+ * regenerated/reordered plan without re-executing what it already has.
+ */
```
pyramation, line 10: So this is entirely additive to our current diff pipeline? Does it modify it?

PR #1806, opened 2026-09-04
feat(graphql-server): authenticate /flush, bound the system lane, name the introspection role

pyramation:
## Summary

Three trust-boundary fixes in the GraphQL server lane (constructive-planning #1611, #1610, #1612 step 1). #1609's header-trusted paths are untouched — its deliverable is an inventory comment on the issue.

**#1611 — `/flush` was unauthenticated.** `flush` becomes a factory so the configured secret is closed over at mount time, and cache invalidation happens only behind a bearer check. No token configured means the route does not exist:

```ts
export const createFlushMiddleware = (opts: ConstructiveOptions): RequestHandler =>
  async (req, res, next) => {
    if (req.url !== '/flush') return next();
    if (!opts.api?.flushToken) return res.status(404).send('Not Found');   // closed by default
    if (!tokensMatch(bearerToken(req), expected)) return res.status(401).send('Unauthorized');
    graphileCache.delete(req.svc_key); svcCache.delete(req.svc_key); ...
  };
```

`tokensMatch` compares sha256 digests with `timingSafeEqual`, so equal-length buffers regardless of the presented token's length. New config: `api.flushToken` / `API_FLUSH_TOKEN` (no default).

**#1610 — the system lane ran as the connecting superuser.** `withPgClient(null, cb)` gives a client with no settings, so `cb` ran as whatever the pool connected as (`postgres`, which bypasses RLS) with each statement in its own implicit transaction. New `withSystemLaneClient` in `pg-query-context` names the role instead, following the `PublicKeySignature.ts` shape — explicit `BEGIN`, transaction-local role, `COMMIT`, `ROLLBACK` + rethrow:

```ts
return withPgClient(null, async (client) => {
  await client.query({ text: 'BEGIN' });
  try {
    await client.query({ text: 'SELECT set_config($1, $2, true)', values: ['role', role] });
    const result = await cb(client);
    await client.query({ text: 'COMMIT' });
    return result;
  } catch (err) {
    try { await client.query({ text: 'ROLLBACK' }); } catch (e) { log.error(...); }
    throw err;   // a broken connection makes ROLLBACK throw too; the real failure still wins
  }
});
```

The role is transaction-local (`set_config(..., true)`), so it is discarded with the transaction and cannot leak to the next borrower of the pooled connection the way a session-level `SET ROLE` would. Converted call sites: `graphile-i18n` translation lookup, and the presigned-url plugin's storage-module lookups (`plugin.ts` provisioning + both delete paths, `download-url-field.ts`, `managed-upload.ts`). Request-scoped `withPgClient(pgSettings, ...)` calls are left alone — including `graphile-bucket-provisioner-plugin`, whose `provisionBucket` mutation already runs with request settings.

The lane's default role is `administrator` (`SYSTEM_LANE_ROLE`). Stated plainly, in the code comment as well as here: `ROLES.md` documents `administrator` as the **BYPASSRLS** role, so **the lane still bypasses RLS**. What changes is that it no longer inherits the *connection's* superuser and its work is one explicit transaction. Narrowing it to an RLS-subject role needs grants that do not exist yet (#1612 phase 3) — verified on the provisioned database: `administrator` holds no direct grants of its own and inherits `anonymous` + `authenticated`.

**#1612 step 1 — introspection ran as superuser.** `makePgService` now receives `pgSettingsForIntrospection` when a role is configured:

```ts
makePgService({ pool, schemas, ...(introspectionRole && { pgSettingsForIntrospection: { role: introspectionRole } }) })
```

New config: `api.introspectionRole` / `API_INTROSPECTION_ROLE`, **unset by default — opt-in**, because of the SDL evidence below.

### SDL evidence: introspecting as `administrator` shrinks the schema

Provisioned a real database (`sdl_evidence`, `constructiveio/postgres-plus:18`, full platform deploy + `pgpm admin-users bootstrap/add`), built the schema twice through `ConstructivePreset` + `makePgService` — once with no introspection role (pool's connecting role) and once with `pgSettingsForIntrospection: { role: 'administrator' }` — and diffed the printed SDL. 57,871 vs. 57,864 lines; the diff is **not empty**:

```diff
 input TablePatch {
-  """Declarative step-up auth guard: ..."""
-  stepUp: JSON

 input TriggerPatch {
-  category: ObjectCategory
-  tags: [String]
```

Three fields vanish, and they vanish as "cannot query field" rather than an authorization error. Cause is grant sizing, not the mechanism: `administrator` lacks `UPDATE` on `metaschema_public.table.step_up` and on `metaschema_public.trigger.category` / `.tags` (it has `SELECT` via `authenticated`, so the read fields survive and only the patch inputs disappear), while the connecting role has it.

So the default flips to unset, per your instruction, and this delta is the concrete grant list #1612 phase 3 has to close before naming a role is safe. `graphql/types` and the server README now carry that warning next to the option.

### `/flush` callers (item 3)

`API_FLUSH_TOKEN` is now load-bearing: **without it the route does not exist (404)**, so any operator who flushes the schema cache over HTTP must set it on the server and send it from every caller.

Swept `constructive-db` for deployed callers — application code, reconciler, `fun`/CLI commands, k8s manifests, Traefik/ingress config, operator docs: **no literal `/flush` reference anywhere in the repo**, so nothing on that side needs token wiring in this PR. `flushService`'s LISTEN/NOTIFY invalidation is a separate path and is untouched, which is how migrations actually invalidate caches today.

Two callers do exist outside these repos, in `diligence-app-demo`'s provisioning package (`packages/provision/src/create-db.ts`, `packages/provision/src/provision.ts`), both `POST`ing to `http://…:3000/flush` with no `Authorization` header. They point at a local dev server and one already tolerates a non-OK response (`Flush returned ${status}`); they need the header added when they next run against a server that sets the token.

### Not fixed here

`graphql/explorer/src/server.ts` has its own unauthenticated `/flush` that deletes from `graphileCache`. Left alone deliberately and reported on #1611.

Test-harness changes worth noting: the presigned-url `fakeDb` throws on unhandled statements by design, so it now answers `BEGIN`/`COMMIT`/`ROLLBACK` as transaction control rather than data queries; the i18n test seed creates `administrator` and grants it schema usage, since lookups now run under that role.

Link to Devin session: https://app.devin.ai/sessions/61d21751dc724d29acb279af0b38cd05
Open in Devin Desktop: https://app.devin.ai/desktop/session/61d21751dc724d29acb279af0b38cd05?variant=devin
Requested by: @pyramation


graphile/graphile-i18n/src/__tests__/setup.sql
```diff
@@ -39,3 +39,14 @@ SELECT setval('i18n_test.posts_id_seq', 10);
 -- Grant access so withPgClient pool can query
 GRANT ALL ON ALL TABLES IN SCHEMA i18n_test TO PUBLIC;
 GRANT ALL ON ALL SEQUENCES IN SCHEMA i18n_test TO PUBLIC;
+
+-- Translation lookups run in the system lane, which sets role=administrator
+-- for the transaction (see pg-query-context's withSystemLaneClient).
+DO $$
+BEGIN
+  CREATE ROLE administrator;
+EXCEPTION WHEN duplicate_object THEN
```
pyramation: why would you create this? isn't this already done in our tests? and then we do it in prod. think this is not necessary.

graphile/graphile-settings/__tests__/upload-resolver.test.ts
```diff
@@ -102,6 +102,8 @@ function fakeContext(opts: { isPublic?: boolean; existingFile?: boolean; failIns
   const client = {
     async query(o: { text: string; values?: unknown[] }) {
       queries.push(o);
+      // Transaction control is not a data query, so it needs no handler.
+      if (/^(BEGIN|COMMIT|ROLLBACK)$/.test(o.text)) return { rows: [] };
```
pyramation: wait, so — what is going on here? why is this here? 

I don't see why we're using regular expressions to check if something is a transaction. That's sloppy, esp. since BEGIN and COMMIT can occur inside the body of a plpgsql function.

pyramation (commented):
Hey, I'm just curious: why did you choose to put this stuff in that package? That package pg query context was supposed to be generic — please explain this pr.

PR #1845, opened 2026-09-17
feat(pgpm): init adds each new module to the workspace CI matrix, sorted

pyramation:
## Summary

The generated workspace CI keeps a **literal, hand-editable** test matrix; `pgpm init` (module) now maintains it instead of the matrix being computed at CI time with `pgpm ls` (the "magic" objected to in pgpm-boilerplates#53/#55).

After a module is scaffolded (pgpm *and* pnpm/lerna/npm workspaces), init calls the new `addToCiMatrix(workspacePath, relativeModulePath)` from `@pgpmjs/core`:

- scans `<workspace>/.github/workflows/*.y(a)ml`
- parses each file with the `yaml` package (new dep on `@pgpmjs/core`) and resolves `jobs.<job>.strategy.matrix.package` for every job — a `package:` key anywhere else (a step's `with:`, `env:`) is never a match
- for each such YAML sequence of plain strings: inserts the module's workspace-relative path (`packages/<name>`), dedupes, sorts with `localeCompare`, and splices the re-rendered sequence into **its own `range` in the source**; the rest of the file is passed through byte-for-byte, so comments, quoting and indentation survive (no `stringify` round-trip)
- keeps the sequence's style — flow `[a, b]` (single- *or* multi-line) or block `- a` — existing items' source text as written, and each item's own comments
- no-op when the entry is already present, the file has no such matrix, the matrix isn't a plain string list (`${{ fromJSON(...) }}`), or the YAML doesn't parse
- prints `Added <mod> to the CI matrix in .github/workflows/ci.yml` per changed file
- best-effort: any fs error (unreadable/unwritable workflow, `.github/workflows` being a file) skips that file/dir and never fails `pgpm init` — the module is already scaffolded by then, so a CI-file problem must not blow up the command

```yaml
# before                         # after `pgpm init` → alpha
package: []                      package: [packages/alpha]
package:                         package:
  - packages/beta                  - packages/alpha
  - packages/delta                 - packages/beta
                                   - packages/delta
```

### Hand-edited arrays keep their comments

A user's matrix is theirs to edit, so per-item comments travel with their item through the sort rather than being dropped:

```yaml
package:                          package:
  - packages/beta   # api    →      - packages/alpha
  # the web app                     - packages/beta   # api
  - packages/delta                  # the web app
                                    - packages/delta
```

`commentBefore`/`comment` are read off each scalar and re-emitted with it; when the first item carries a leading comment block the replaced span is extended back over those lines so they can't be duplicated. A flow sequence containing comments is left untouched instead (re-rendering comments inside `[...]` isn't worth it).

Backstop, so no shape can silently eat a comment: before an edit is accepted, the candidate result is re-parsed and its comment multiset compared with the original document's (`commentsOf`, via `yaml`'s `visit`). Any loss, duplication, or parse failure → that matrix is left exactly as it was.

Companion boilerplate change (ships `package: []` + `cd ./${{ matrix.package }}`): constructive-io/pgpm-boilerplates#56. Note GitHub rejects an empty literal matrix, so a workspace with zero modules has an invalid `ci.yml` until its first `pgpm init`; `pgpm init -w` produces a valid one immediately.

Tests: `pgpm/core/__tests__/core/ci-matrix.test.ts` — whole-file equality for flow/block/empty/quoted/multi-job/unrelated-key/expression/unparseable/CRLF/anchor, plus a table-driven case asserting the comment multiset is identical before and after across 25 hand-edited shapes (extra spaces, trailing comments, comment blocks above items, multi-line flow, `#` inside quoted values). E2E in `pgpm/cli/__tests__/init.test.ts` inits `zeta` then `alpha` and asserts `package: [packages/alpha, packages/zeta]` with the surrounding comment preserved.

Link to Devin session: https://app.devin.ai/sessions/dd4055c50bcf4b1e8e604211038aaa82
Open in Devin Desktop: https://app.devin.ai/desktop/session/dd4055c50bcf4b1e8e604211038aaa82?variant=devin
Requested by: @pyramation


pgpm/core/src/core/ci-matrix.ts
```diff
@@ -0,0 +1,97 @@
...
+  [...new Set(entries)].sort((left, right) => left.localeCompare(right));
+
+/**
+ * Add a module to the `package:` matrix of a workspace's CI workflows, keeping
+ * the list sorted. The list stays a plain, hand-editable array: workflows
+ * without one, or without the key, are left alone.
+ *
+ * Returns the workflow files that changed, relative to `workspacePath`.
+ */
+export const addToCiMatrix = (
```
pyramation, line 252: explain how this works. explain also how we read/write yaml and ensure precision.

pgpm/core/src/core/ci-matrix.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Insert `entry` into the first `package:` sequence of a workflow, in either
+ * flow (`package: [a, b]`) or block form, preserving indentation, comments and
+ * everything else in the file.
+ */
+export const addToMatrixYaml = (source: string, entry: string): string => {
+  const lines = source.split('\n');
+
+  for (let index = 0; index < lines.length; index += 1) {
+    const flow = lines[index].match(
+      new RegExp(`^(\\s*)${MATRIX_KEY}:\\s*\\[(.*)\\]\\s*$`)
```
pyramation: how do we know it's the matrix key and not some other key for another workflow?

PR #1852, opened 2026-09-21
F17: bind Graphile reuse to the exact build contract

Zetazzz:
Implements F17 from constructive-io/constructive-planning#1693, stacked on #1747 (F04).

A logical service key can resolve to different databases, schemas, roles, plugin settings, or compute bindings. Server and Explorer snapshot their effective build inputs and use a domain-separated, process-local HMAC fingerprint for cache reuse. Functions and opaque objects retain reference identity; credentials stay out of cache keys. The captured inputs also construct the preset.

Service, database, and physical-pool invalidation use explicit entry metadata to remove every build variant. Public schema notifications invalidate by database before hostname lookup. Existing scoped-introspection wiring is preserved.

The server reuses the API, pool, service key, module loader, and PostgreSQL settings owned by `req.constructive`. Missing request claims were added to the express-context owner, including current request protection, principal/read-only credentials, trusted private identity headers, and anonymous database attribution. Shared schemas read these settings per request. The canonical errors package retains internal causes and registered HTTP status/code while excluding private messages and context. `/flush` authenticates and invalidates before starting a Graphile build.

Validation: 130 express-context tests; 50 graphile-cache tests at this layer; targeted error/snapshot/response/observability tests; seven real PostgreSQL/scoped-routing HTTP scenarios replace the mocked middleware tests. Server, Explorer, context and error-package builds passed. Final stack validation is recorded in #1855.

Dependency stack: #1747 → #1852 (F17) → #1853 (F19) → #1854 (F20) → #1855 (F21). F20 logically depends on F04/F17 and is stacked after F19 to share the admitted publication owner.


Zetazzz:
Followed up on the three observations in the approval summary:

- Fixed the missing-service-key case in 66f1caa55dbff49fffd745963f02d0266333bc08. `undefined`, `null`, and empty-string keys now return without invalidating entries. Three regressions preserve both tagged and legacy entries. The guard is also propagated ahead of flight invalidation throughout #1853, #1854 and #1855.
- The flush/publication race is addressed by the dependent F20 PR #1854: invalidated generations are fenced before publication, including builds already in flight.
- Kept compute resolution before cache lookup because its effective configuration is part of the exact build identity. Skipping it without an authoritative configuration revision could reuse a schema built with stale compute settings.

Validation: 6 focused invalidation tests on this branch; the final dependency stack passes all 84 graphile-cache tests, all 8 server flush tests, and the graphile-cache build.

